PR #3, opened 2017-01-15
iocage readme changes

jungle-boogie:
first pass of changes for readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Start the jail:
 `iocage start myjail`
 
 **REQUIREMENTS**
-- FreeBSD 9.3-RELEASE amd64 or newer
+- FreeBSD 10.3-RELEASE amd64 or newer
 - ZFS file system
 - Python 2.7
 - Optional - Kernel compiled with:
--- a/README.md
+++ b/README.md
@@ -7,10 +7,11 @@ iocage is a jail/container manager amalgamating some of the best features and
 technologies the FreeBSD operating system has to offer. It is geared for ease
  of use with a simple and easy to understand command syntax.
 
-iocage is in the FreeBSD ports tree as sysutils/iocage.
-To install using binary packages, simply run: `pkg install iocage`
+iocage is in the FreeBSD ports tree as sysutils/py-iocage.
+To install using binary packages, simply run: `sysutils/py-iocage`
 
-If cloning the repo directly, cd into the cloned directory and run `pip install .` as root.
+If cloning the repo directly, ensure you have python 2.7,
+cd into the cloned directory and run `pip install .` as root.
 
 ## WARNING:
 - This is beta quality software, there be dragons! Please report them.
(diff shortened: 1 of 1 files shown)


skarekrow:
Hello!

Thanks for the PR, but a few things aren't going to be merged from it. First, the port will be named `iocage`, I was waiting for @kmoore134 to make a port for me. @bartekrutkowski was just incredibly quick on his feet (appreciate that robak!) and already made one. But I don't expect that to stay as I don't want that naming scheme for iocage. The current `iocage` will become `iocage_legacy` and this will become `iocage` with `iocage_legacy` conflicting with it.

For the final edit, why remove 9.3? Even if it is EOL, I'm stating that 9.3 and up is required, and that remains true whether or not upstream supports it. Many users may stick with it for some time.

I *will* however take the Python 2.7 requirement, d'oh! Good catch. If you could update and then squash this PR with just that one addition, I'll merge this :)

bartekrutkowski:
@skarekrow Regarding the port name, I am afraid software name doesn't get a priority over the port name. New iocage version is not only based on Python, what requires `py-` prefix in Ports Tree, but also has 'incompatible' version number (0.9 currently is lower than `sysutils/iocage` 1.7.6 version number) with existing iocage port, so I am afraid you will have to live with `sysutils/py-iocage` in Ports Tree. I've already updated `sysutils/iocage` port to use the `iocage_legacy` repository.

README.md
```diff
@@ -7,10 +7,11 @@ iocage is a jail/container manager amalgamating some of the best features and
 technologies the FreeBSD operating system has to offer. It is geared for ease
  of use with a simple and easy to understand command syntax.
 
-iocage is in the FreeBSD ports tree as sysutils/iocage.
-To install using binary packages, simply run: `pkg install iocage`
+iocage is in the FreeBSD ports tree as sysutils/py-iocage.
+To install using binary packages, simply run: `sysutils/py-iocage`
```
bartekrutkowski, line 11: This line is incorrect, it should say `pkg install py27-iocage`.

jungle-boogie: Hi @bartekrutkowski,

But it looks like what's in my diff is correct:
https://www.freshports.org/sysutils/py-iocage/

jungle-boogie:
Hi @skarekrow,

I'll work with @bartekrutkowski and determine the correct port/package name and open a new PR then.

Thanks!

README.md
skarekrow: @jungle-boogie If you look at PKGNAME you will see it's py27-iocage.

@bartekrutkowski We still need a conflict on these, I don't intend for them to coexist.

jungle-boogie: @skarekrow,

Yeah, what's up with that. https://www.freshports.org/sysutils/py-iocage but says pkg install py27-iocage

**doesn't exist** https://svnweb.freebsd.org/ports/head/sysutils/py27-iocage 
**exists** https://svnweb.freebsd.org/ports/head/sysutils/py-iocage

skarekrow: Yeah all python ports do that so we can have Python 2 and 3 ports coexist. It's a fair trade off. I just committed and gave you credit in title. Thanks again!

PR #22, opened 2017-01-24
Use safer click.confirm() to confirm with user to proceed with an opp…

drewja:
…eration. Remove an un-used variable count from destroy.py->destroy_cmd(). Add file .gitignore.


skarekrow:
Nice! Thank you

PR #30, opened 2017-02-03
Fix incorrect or operator usage.

cpforbes:
This fixes a few improper uses of the or operator.


skarekrow:
Thanks! Just a few questions as a python noob, the usages you're replacing do function correctly. So are they just not pythonic or is there something else I'm not aware of?

skarekrow:
Well that's funny, I must not be remembering, it totally did not work. Thanks again for this PR.

Issue #62, opened 2017-03-07
Missing IPv6 support for VNET


joachimmathes:
I was wondering why my `ip6_addr` and `defaultrouter6` settings did not show any effect. But after looking at https://github.com/iocage/iocage/blob/master/iocage/lib/ioc_start.py#L298 it seems that IPv6 is not supported, yet. Am I right, or did I miss something?

Best
Jo

skarekrow:
Depends on your version, but `ip6_addr` works correctly with `master`. `defaultrouter6` does not as it requires VNET. That code block you pasted is for VNET, and yes I haven't added IP6 VNET support yet, so you're going to have to use the normal shared IP, or IP4 addresses in the meantime. Sorry about that!

joachimmathes:
VNET IPv6 is actually what we need. Do you already have plans, when you are going to implement this feature? If you want, I can offer you support with that and try to implement it.

skarekrow:
Ah, sorry!

Yes I will have it done before 1.0, which I don't have an exact timetable on, but I'd think I would have this finished before April. Feel free to take a stab at it though, I think it will largely be a copy paste of the ip4 VNET goop, which would make it an ideal candidate to split that function up so we can reuse as much as we can for both 6 and 4.

I don't think it will be a terribly hard task, but you never know :D

joachimmathes:
I'll give it a try, but don't hesitate, if you consider to get this done. ;)

joachimmathes:
I am not able to provide a pull request, yet. But - as a proof of concept - these modifications in `ioc_start.py` make VNET IPv6 work:
```diff
301,302c301,302
<             ip4_addr = self.get("ip6_addr")
<             defaultgw = self.get("defaultrouter6")
---
>             ip4_addr = self.get("ip4_addr")
>             defaultgw = self.get("defaultrouter4")
373c373
<                                      "ifconfig", iface, "inet6",ip4, "up"],
---
>                                      "ifconfig", iface, ip4, "up"],
376c376
<                                     self.uuid), "route", "add", "-6", "default",
---
>                                     self.uuid), "route", "add", "default",
```

skarekrow:
Great news, then it should be an easy enough modification

skarekrow:
Working on this now, give me a few :)

skarekrow:
Alright that took a bit to get a vnet env up and running, please test and let me know 👍

joachimmathes:
Thank you!
[...]

PR #63, opened 2017-03-08
Allow VNET IPv6 only jails

joachimmathes:
This pull request refers to issue https://github.com/iocage/iocage/issues/62

* It allows VNET IPv6 only jails.
* It splits logic into two methods to clean code and stick to 80 characters limit.


skarekrow:
I like it! Thanks :)

Issue #65, opened 2017-03-15
Allow setting a specific jail uuid when creating a jail


Licenser:
# Short

It would be nice to have a option like:

```
iocage create uuid=<put my uuid here> ...
```

This would then allow instead of doing this:

https://github.com/iocage/iocage/blob/master/iocage/lib/ioc_create.py#L48

using the supplied UUID.

# Long

I'm trying to port parts of [Project-FiFo](https://project-fifo.net) over to FreeBSD and currently am using iocage to interact with the jails. FiFo generates a UUID before starting the zone (or jail) creation process to allow tracing and displaying things in an asyncronous manner. This is problematic when the `create` command can't enforce a specific UUID for the zone. The current approach I found is using the tag 'fifo:<fifo-uuid>' which in itself is slightly problematic as it makes the output of `iocage list` less human readable.

skarekrow:
Closed by #68

PR #68, opened 2017-03-15
Custom uuid

Licenser:
This PR adds a `--uuid (-u)` option to the `iocage create` command that allows specifying the uuid used for the jail. This is related to #65 .

### DIFF
--- a/iocage/cli/create.py
+++ b/iocage/cli/create.py
@@ -43,8 +43,8 @@ def create_cmd(release, template, count, props, pkglist, basejail, empty,
 
     if short and uuid:
         raise RuntimeError(
-            "Can't use --short (-s) and --uuid (-u) at the same time!!")
-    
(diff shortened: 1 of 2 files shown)


Licenser:
FWIW the travis check didn't fail for code reasons but seems to be a bad travis config

skarekrow:
Looks good except you should make sure that UUID isn't used elsewhere. If the user supplies the same UUID twice, hilarity *will* ensue as we count on that being unique.

Otherwise nice PR! 👍

Licenser:
That's an excellent point :) totally overlooked that. I'll update that later today.

Licenser:
It's checking and erroring out when the UUID exists. The test is performed either way with a provided or generated UUID, seems like OK to always check.

skarekrow:
LGTM! I agree, it doesn't matter if we test for all.

Thanks :)

PR #77, opened 2017-03-21
pcpu is an number not on/off

### LINKED ISSUE #76: pcpu values mismatch between implementation and docs
according to the docs [1] pcpu follows the same rules as memory, however the implementation expects `on` or `off`.

```
ERROR: 100: deny is not a valid value for pcpu.
Value must be off or on
```

http://iocage.readthedocs.io/en/latest/advanced-use.html#limit-cpu-use-by

Licenser:
Might resolve #76 at least it doesn't error and behaves related to the docs. I'm not sure how the whole resource thing ties together so it might actually do nothing at all other then save the value somewhere.


skarekrow:
The change in `ioc_create` isn't needed as the default would be fine. Otherwise this should be good! We may want to actually check for a range of numbers and then `deny, log, etc`. I'm not sure how many of the actions `pcpu` supports from the manpage.

Licenser:
I just made those up, can you give me a hunt what man page it's documented it? :D

skarekrow:
Sure :D It's in `rctl(8)`

Licenser:
don't think we can really check it it's % per core (which I personally like ;) so 100% == 1 core. but since cores could change checking it might not make much sense.

skarekrow:
Yeah I think it's best we handle this simple (how your change is) and if something happens add more logic.

Licenser:
Oh rebarding the default, should all values be 'off' then? also the memory? I thought `off` here means 'not implemented'.

skarekrow:
Ah yes they should, good catch!

skarekrow:
Thanks again :+1:

PR #97, opened 2017-04-07
- Create a generic module to be able to avoid some duplicate code.

araujobsd:
* libiocage aims to be a bridge between cli and lib, it is in its initial stage and it can be considered as an experiment/prototype.


skarekrow:
👍

Issue #110, opened 2017-04-20
Documentation for upgrading iocage


ketralnis:
As far as I can tell, there is currently no documentation or how-to for migrating iocage from the older shell script based system to the new Python based system. This leaves users of the older version in an odd state.

What needs documenting:

- [ ] How to update the Ports version. For instance:

```sh
$ portversion -v iocage
iocage-1.7.6                ?  error - origin not found

$ grep VERSION /usr/ports/sysutils/py-iocage/Makefile
PORTVERSION=    0.9.5
```

- [ ] How to migrate the jails themselves. It appears that the file system layout has changed, so will simply removing iocage and installing py-iocage work as expected?

- [ ] What to do if you currently have basejails, which are no longer supported (see also https://github.com/iocage/iocage/issues/50)

And probably other stuff I'm forgetting

skarekrow:
Indeed there is documentation needed. But the good news is yes all you have to do if you were using the latest stable shell version of iocage (1.7.6) is remove iocage, install py-iocage and it will automatically do all the migration for you.

I prefer to run `iocage list` as root to do that for me. If you use `iocage-legacy`'s `develop` branch, then you will have to run `iocage migrate` once you have installed `py-iocage` and removed `iocage`.

Please don't use the `py-iocage` port, it is old and very out of date. Use `master` from GitHub, or the `py3-iocage` port.

ketralnis:
> install `py-iocage` and it will automatically do all the migration for you. [...] Please don't use the `py-iocage` port, it is old and very out of date. Use master from GitHub, or the py3-iocage port.

Ouch, that absolutely would have been my first choice because I rely on python27 and most of the py27/py3 ports are the same code with just different python version dependencies.

That should probably be passed on to the ports maintainer (which looks like robak@FreeBSD.org) so that the port can just be deleted if it's not valid anymore.

ketralnis:
> remove iocage, install py-iocage and it will automatically do all the migration for you.

Is that true even if the old jails are basejails?

skarekrow:
> Is that true even if the old jails are basejails?

Yes, it should migrate those as well. If it doesn't, please make a ticket. So backups should be done first :)

> Ouch, that absolutely would have been my first choice because I rely on python27 and most of the py27/py3 ports are the same code with just different python version dependencies.<br>
That should probably be passed on to the ports maintainer (which looks like robak@FreeBSD.org) so that the port can just be deleted if it's not valid anymore.

[...]

PR #111, opened 2017-04-20
Add documentation for upgrading from iocage_legacy

ketralnis:
Adds documentation to README.md for upgrading from the older shell-based `iocage_legacy` to the newer Python-based `iocage`.

This documentation is not added to readthedocs under the theory that this is a one-time operation only necessary for a small number of users so it's probably not needed in the reference docs. But I'd be happy to clone it there if you think it's useful

Fixes https://github.com/iocage/iocage/issues/110

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,14 +45,11 @@ To install subsequent updates: `git pull` and `pip3.6 install -U .`
 
 This repository replaces `iocage_legacy`. To upgrade to the current version:
 
-1. Back up your data
-2. Stop the jails (`service iocage stop`)
+1. Stop the jails (`service iocage stop; iocage stop ALL`)
+2. Back up your data
 3. Remove the old `iocage` package if it is installed (`pkg delete iocage`)
 4. Install `py3-iocage` using one of the methods above
-5. Migrate the jails. This can be done by running `iocage list` as root. If you
-   used `iocage-legacy`'s `develop` branch, then you will have to run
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,6 +41,19 @@ To install subsequent updates: `git pull` and `pip3.6 install -U .`
...
+This repository replaces `iocage_legacy`. To upgrade to the current version:
+
+1. Back up your data
+2. Stop the jails (`service iocage stop`)
+3. Remove the old `iocage` package if it is installed (`pkg delete iocage`)
+4. Install `py3-iocage` using one of the methods above
+5. Migrate the jails. This can be done by running `iocage list` as root. If you
+   used `iocage-legacy`'s `develop` branch, then you will have to run
+   `iocage migrate` first. If you previously had basejails, they will be
+   promoted to ____
```
ketralnis: I don't know the answer to this :)

skarekrow:
Ok I'll answer here so you can add it all in the PR. 

Basejails remain as they are for `master`/stable users. You will still have all the cloned datasets. The `develop` guys get migrated to a normal clone jail. I can't recall if I migrate their basejails or not off the top of my head. So you can omit the `develop` basejails portion and I can add it later. 

I would also add a `iocage stop ALL` command after `service iocage stop` as users may have jails they don't use boot` on. I'd also have the backup direction after stopping the jails.

skarekrow:
Also thank you for the PR!

Issue #202, opened 2017-06-23
feature request: iocage commands making guesses about short UUIDs need autoconfirm option


igalic:
The following code (in [iocage/cli/create.py](https://github.com/iocage/iocage/blob/master/iocage/cli/create.py#L100))
```
        except ValueError:
            ioc_common.logit({
                "level"  : "WARNING",
                "message": "This may be a short UUID, template creation only"
                           " supports TAGs"
            })
            if not click.confirm("\nProceed?"):
                exit()
```
makes it tricky to automate jail creation… I don't know how iocage guesses that this:

```
iocage create --template=f11-ats6 tag=cdn ip4_addr="vtnet0|172.16.0.7/12" ip6_addr="vtnet0|2a03:b0c0:3:d0::4c97:6007"
This may be a short UUID, template creation only supports TAGs

Proceed? [y/N]:
```

looks like a short-UUID, but here we are…

(before you suggest using `yes y | iocage…` i'd rather not have to exec-pipe in ruby;)

skarekrow:
Fixed by #203

PR #203, opened 2017-06-23
add auto-confirm option for `create` subcommand

igalic:
this pr addresses #202

### DIFF
--- a/iocage/cli/create.py
+++ b/iocage/cli/create.py
@@ -70,7 +70,7 @@ def validate_count(ctx, param, value):
               default=False, is_flag=True)
 @click.argument("props", nargs=-1)
 def cli(release, template, count, props, pkglist, basejail, empty, short,
-        name, _uuid):
+        name, _uuid, force):
     if name:
         _props = []
         if f"tag={name}" not in props:
@@ -102,9 +102,9 @@ def cli(release, template, count, props, pkglist, basejail, empty, short,
         except ValueError:
             if not force:
                 ioc_common.logit({
-                    "level"  : "WARNING",
-                    "message": "This may be a short UUID, template creation only"
-                            " supports TAGs"
+                    "level":   "WARNING",
+                    "message": "This may be a short UUID, "
+                               "template creation only supports TAGs"
                 })
                 if not click.confirm("\nProceed?"):
                     exit()
--- a/iocage/cli/create.py
+++ b/iocage/cli/create.py
@@ -66,6 +66,8 @@ def validate_count(ctx, param, value):
 @click.option("--short", "-s", is_flag=True, default=False,
               help="Use a short UUID of 8 characters instead of the default "
                    "36")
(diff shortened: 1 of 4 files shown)


iocage/cli/create.py
```diff
@@ -98,13 +100,14 @@ def cli(release, template, count, props, pkglist, basejail, empty, short,
...
             })
         except ValueError:
-            ioc_common.logit({
-                "level"  : "WARNING",
-                "message": "This may be a short UUID, template creation only"
-                           " supports TAGs"
-            })
-            if not click.confirm("\nProceed?"):
-                exit()
+            if not force:
```
gronke, line 103: force is not defined yet. Did you forget to put in in the cli method params?

iocage/cli/create.py
```diff
@@ -98,13 +100,14 @@ def cli(release, template, count, props, pkglist, basejail, empty, short,
...
+                ioc_common.logit({
+                    "level"  : "WARNING",
+                    "message": "This may be a short UUID, template creation only"
```
gronke: The linter requires lines to be max 80 characters long. I suggest to shift a bit more of the text to the next line.

iocage/cli/create.py
```diff
@@ -98,13 +100,14 @@ def cli(release, template, count, props, pkglist, basejail, empty, short,
...
+                            " supports TAGs"
```
gronke: From Tavis CI:
```
+./iocage/cli/create.py:107:29: E131 continuation line unaligned for hanging indent
```

Seems this line needs to be indented on a different level.

gronke (commented):
👍 besides the few things the linter complains about :)

iocage/cli/create.py
igalic: my personal preference is to have error messages in-tact, to make grepping for them easier
failing that, i think the next best would be to break on `,` in this case.

iocage/cli/create.py
igalic: yeeeaah… i did… i was looking for where this is passed, and decided, it's done by magic.
thanks for the clarification ;)

iocage/cli/create.py
skarekrow: Breaking on `,` in this case is fine

skarekrow:
Once you fix the things @gronke (<3) reviewed, I'll merge it in :)

igalic:
i'm still struggling with my dev setup here… last bit seems to be [yapf's lack for 3.6 support](https://github.com/google/yapf#python-versions)

skarekrow (approved):
Changing the `:` in this isn't the issue ;) My editor will auto move these back as I like them to be symmetrical.

The flake issue is that there are multiple spaces *after* not before the `:`, but this is something I can easily fix so no worries.

skarekrow:
Thanks again @igalic 👍

PR #209, opened 2017-06-26
Fix mac_b address construction

joachimmathes:
mac_b seems to be invalid as its prefix is `2f:f6`. Due to a missing leading zero it's not a valid _invidiual_ and _local_ MAC. Don't know if `zfill` is a python best practice. I looked it up in former iocage code. :smile:

### DIFF
--- a/iocage/lib/ioc_start.py
+++ b/iocage/lib/ioc_start.py
@@ -580,7 +580,7 @@ class IOCStart(object):
 
     def __generate_mac_address_pair(self, nic):
         mac_a = self.__generate_mac_bytes(nic)
-        mac_b = hex(int(mac_a, 16) + 1)[2:]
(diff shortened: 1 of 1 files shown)


skarekrow:
Haha I don't either. But it works right? ;)

joachimmathes:
Sorry, I forgot to rename the branch, since I thought the static `mac_prefix` was the problem. But speaking of this: It is static at the moment, isn't it? (https://github.com/iocage/iocage/blob/master/iocage/lib/ioc_start.py#L578) 
Is this due to the hashed and therefore unique MAC suffix? Or will it be configurable again in the future?

skarekrow:
Yeah that's static currently, the property does exist, it was just lost in the new mac generation, I restored it just now in 3dfd6b7a93d7c1b727ddd2d124e9c8f0c71fcb24 👍 good eye!

PR #226, opened 2017-06-28
Add help text to create.py

Mrt134:
Added help text to the -r, -t, -b, and -e options.
Rework the -f option to match the other options' syntax.
TODO - need clarification on the -c and -p options before adding that help text.

- [x] Explain the feature

### DIFF
--- a/iocage/cli/create.py
+++ b/iocage/cli/create.py
@@ -53,21 +53,28 @@ def validate_count(ctx, param, value):
 
 @click.command(name="create", help="Create a jail.")
 @click.option("--count", "-c", callback=validate_count, default="1")
-@click.option("--release", "-r", required=False)
-@click.option("--template", "-t", required=False)
+@click.option("--release", "-r", required=False,
+              help="Specify the release to use for the new jail.")
+@click.option("--template", "-t", required=False,
+              help="Flag this jail as a template, which allows for rapid "
+                   "redeployment of a customized jail.")
 @click.option("--pkglist", "-p", default=None)
 @click.option("--name", "-n", default=None,
               help="Provide a specific name and tag instead of an UUID for"
-                   " this jail")
+                   " this jail.")
 @click.option("--uuid", "-u", "_uuid", default=None,
-              help="Provide a specific UUID for this jail")
-@click.option("--basejail", "-b", is_flag=True, default=False)
-@click.option("--empty", "-e", is_flag=True, default=False)
+              help="Provide a specific UUID for this jail.")
+@click.option("--basejail", "-b", is_flag=True, default=False,
+              help="Set the new jail type to a basejail. Basejails "
+                   "re-clone their base filesystems at each startup.")
+@click.option("--empty", "-e", is_flag=True, default=False,
+              help="Create an empty jail used for unsupported or custom "
+                   "jails.")
 @click.option("--short", "-s", is_flag=True, default=False,
               help="Use a short UUID of 8 characters instead of the default "
-                   "36")
-@click.option("--force", "-f", help="Skip the interactive question.",
-              default=False, is_flag=True)
+                   "36.")
+@click.option("--force", "-f", is_flag=True, default=False,
+              help="Skip the interactive question.")
 @click.argument("props", nargs=-1)
 def cli(release, template, count, props, pkglist, basejail, empty, short,
         name, _uuid, force):
--- a/iocage/cli/create.py
+++ b/iocage/cli/create.py
@@ -52,27 +52,32 @@ def validate_count(ctx, param, value):
 
 
 @click.command(name="create", help="Create a jail.")
-@click.option("--count", "-c", callback=validate_count, default="1")
+@click.option("--count", "-c", callback=validate_count, default="1",
+              help="Designate a number of jails to create. Jails are"
+                   " numbered sequentially.")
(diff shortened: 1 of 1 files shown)


iocage/cli/create.py
```diff
@@ -53,21 +53,28 @@ def validate_count(ctx, param, value):
 
 @click.command(name="create", help="Create a jail.")
 @click.option("--count", "-c", callback=validate_count, default="1")
-@click.option("--release", "-r", required=False)
-@click.option("--template", "-t", required=False)
+@click.option("--release", "-r", required=False,
+              help="Specify the release to use for the new jail.")
```
skarekrow: `release` should be all caps `RELEASE` as it's specifically meant to reference the available `RELEASEs`.

iocage/cli/create.py
```diff
@@ -53,21 +53,28 @@ def validate_count(ctx, param, value):
 
 @click.command(name="create", help="Create a jail.")
 @click.option("--count", "-c", callback=validate_count, default="1")
```
skarekrow: `-c` will create that many jails. So if you issue `iocage create -c 20 -r 11.0-RELEASE tag="foo"` you will get 20 jails, labeled `foo_1` to `foo_20`. I'll leave the wording up to you :P

iocage/cli/create.py
```diff
@@ -53,21 +53,28 @@ def validate_count(ctx, param, value):
...
-@click.option("--release", "-r", required=False)
-@click.option("--template", "-t", required=False)
+@click.option("--release", "-r", required=False,
+              help="Specify the release to use for the new jail.")
+@click.option("--template", "-t", required=False,
+              help="Flag this jail as a template, which allows for rapid "
+                   "redeployment of a customized jail.")
 @click.option("--pkglist", "-p", default=None)
```
skarekrow: `-p` allows you to specify a JSON file who's purpose is to install each package inside the jail after creation.

Example:
```
{
    "pkgs": [
    "foo",
    "bar"
    ]
}
``` I have an error prompt for the user if it's in an incorrect format, so perhaps just explaining the concept is enough without also showing the format of the JSON.

File: iocage/cli/create.py [hunk 4]
```diff
@@ -53,21 +53,28 @@ def validate_count(ctx, param, value):
 
 @click.command(name="create", help="Create a jail.")
 @click.option("--count", "-c", callback=validate_count, default="1")
-@click.option("--release", "-r", required=False)
-@click.option("--template", "-t", required=False)
+@click.option("--release", "-r", required=False,
+              help="Specify the release to use for the new jail.")
+@click.option("--template", "-t", required=False,
+              help="Flag this jail as a template, which allows for rapid "
+                   "redeployment of a customized jail.")
```
skarekrow: I prefer to have my spaces on the next line, so this would be:
```
              help="Flag this jail as a template, which allows for rapid"
                   " redeployment of a customized jail.")
```

iocage/cli/create.py
```diff
@@ -53,21 +53,28 @@ def validate_count(ctx, param, value):
...
-                   " this jail")
+                   " this jail.")
 @click.option("--uuid", "-u", "_uuid", default=None,
-              help="Provide a specific UUID for this jail")
-@click.option("--basejail", "-b", is_flag=True, default=False)
-@click.option("--empty", "-e", is_flag=True, default=False)
+              help="Provide a specific UUID for this jail.")
+@click.option("--basejail", "-b", is_flag=True, default=False,
+              help="Set the new jail type to a basejail. Basejails "
+                   "re-clone their base filesystems at each startup.")
```
skarekrow: This is only true for the legacy implementation. The new one actually mounts the specified `RELEASEs` directories as `nullfs` mounts over the jails directories.

So `11.0-RELEASE/root/bin` will be mounted as a R/O nullfs mount over `foo/root/bin`, saving space and making moving to a new `RELEASE` much faster. Once the upgrade mechanism is figured out and implemented that is ;)

iocage/cli/create.py
```diff
@@ -53,21 +53,28 @@ def validate_count(ctx, param, value):
...
+@click.option("--empty", "-e", is_flag=True, default=False,
+              help="Create an empty jail used for unsupported or custom "
+                   "jails.")
```
skarekrow: Same space comment as above.

skarekrow (changes requested):

iocage/cli/create.py
Mrt134: Complete - see new commit on this PR.

iocage/cli/create.py
Mrt134: Complete - see new commit on this PR.

iocage/cli/create.py
Mrt134: Complete - see new commit on this PR.

iocage/cli/create.py
Mrt134: Complete - see new commit on this PR.

iocage/cli/create.py
Mrt134: Complete - see new commit on this PR.

iocage/cli/create.py
Mrt134: Complete - see new commit on this PR.

skarekrow:
Thanks! :+1:

PR #287, opened 2017-07-14
Allow setting ip(4|6)_addr without an interface name

### LINKED ISSUE #193: Allow setting ip4_addr without the interface name
`iocage create` requires `ip4_addr` to contain both an interface name and an IP address, or nothing at all.    I believe it should not require the interface name.

Here's my use case:  I want jails that share the host machine's IP.  The jails need network access, but don't need to accept inbound connections.  I'm in a VPS environment where I can't allocate multiple IPs and I'd like to avoid the overhead and complexity of internal NAT.

The `ip4_addr` property is passed to `jail(8)` as `ip4.addr` on start.  From the `jail(8)` manpage:
> If an interface is given before the IP address, an alias for the address will be added to that interface, as it is with the interface parameter.

If the alias it tries to add is also the host's primary IP, it causes the host to drop off the network and not come back.  The only way to share the host's IP is to not tell `jail(8)` the interface name so that it doesn't try to add an alias.

If I edit a jail's `config.json` so that `ip4_addr` contains only an IP address, everything works as I expect: the jail starts, it has network access, and it shares the IP with the host.

I'm not positive that this is the best way to accomplish my goal, but this is what I used to do with `ezjail`.

rwestlund:
Close #193.

- [x] Explain the feature

### DIFF
--- a/iocage/lib/ioc_json.py
+++ b/iocage/lib/ioc_json.py
@@ -794,10 +794,10 @@ class IOCJson(object):
             "host_domainname"      : ("string",),
             "host_hostname"        : ("string",),
             "exec_fib"             : ("string",),
-            "ip4_addr"             : ("|",),
+            "ip4_addr"             : ("string",),
             "ip4_saddrsel"         : ("0", "1",),
             "ip4"                  : ("new", "inherit", "none"),
-            "ip6_addr"             : ("|",),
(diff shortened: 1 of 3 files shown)


skarekrow:
Looks good, thanks!

PR #291, opened 2017-07-18
ensure templates have correct type on creation

igalic:
when creating templates directly with

    iocage create --name foo template=yes

we also need to ensure that the `type` property is set correctly


skarekrow:
Thanks :D

PR #292, opened 2017-07-19
fix name validation: names should be > 0 char

igalic:
previously, `name` was permitted to be empty (`*`), we now require at
least one character from the allowed set (`+`).


skarekrow:
Thanks ;)

Issue #290, opened 2017-07-18
iocage list -lt shows templates as type `jail`


igalic:
on 44e929c9997b0e27f5794328c4645e933baf43d3, `iocage list -lt` all `type`s are displayed as `jail` — and not as template:

```
+-----+------------+------+-------+------+------------------+----------------------+-----+----------+
| JID |    NAME    | BOOT | STATE | TYPE |     RELEASE      |         IP4          | IP6 | TEMPLATE |
+=====+============+======+=======+======+==================+======================+=====+==========+
| -   | php71      | off  | down  | jail | 11.0-RELEASE-p10 | vtnet0|172.16.0.2/12 | -   | -        |
+-----+------------+------+-------+------+------------------+----------------------+-----+----------+
| -   | static-web | off  | down  | jail | 11.0-RELEASE-p10 | vtnet0|172.16.0.3/12 | -   | -        |
+-----+------------+------+-------+------+------------------+----------------------+-----+----------+
```

this is a regression.

igalic:
apparently, this has been written wrongly in `/iocage/templates/<foo>/config.json`
****
so, from this we can probably guess that it's not so much a `list` issue, but actually a `create` issue.

igalic:
```diff
diff --git a/iocage/lib/ioc_create.py b/iocage/lib/ioc_create.py
index 9b20912..1e87173 100644
--- a/iocage/lib/ioc_create.py
+++ b/iocage/lib/ioc_create.py
@@ -287,6 +287,7 @@ class IOCCreate(object):
                     iocjson.json_write(config)  # Set counts on this.
                     location = location.replace("/jails/", "/templates/")
 
+                    iocjson.json_set_value("type=template")
                     iocjson.json_set_value("template=yes")
 
                 try:
```
i think this should fix it.

igalic:
pr can be found here: https://github.com/iocage/iocage/pull/291

skarekrow:
Nice catch!

skarekrow:
Closed by #291

igalic:
something still seems to be messing with this, but i haven't tracked down yet what

skarekrow:
It's just showing the type of the templates. If you had old ones that had the type of `jail` then it will still be wrong.

igalic:
i don't have any old templates, i've been creating and destroying stuff while developing `puppet-jail`. a lot.

basically, every time i got an exception i couldn't track down, i'd purge the jails/templates zfs'es.

[...]

PR #298, opened 2017-07-25
extract configuration in external functions

igalic:
as a continuation to #293, we use try to extract some infrastructure
setup code from the error handling.

For now, this is happening at the expense of mild duplication, but that
too might be fixed in future patches

### DIFF
--- a/iocage/lib/ioc_create.py
+++ b/iocage/lib/ioc_create.py
@@ -86,6 +86,106 @@ class IOCCreate(object):
             iocage.lib.ioc_destroy.IOCDestroy().destroy_jail(location)
             sys.exit(1)
 
+    def _get_template_config_fn(self, jail_uuid, path):
+        config_fn = iocage.lib.ioc_json.IOCJson(path).json_get_value
+        try:
+            config_fn("cloned_release")
+        except (IOError, OSError, FileNotFoundError, UnboundLocalError):
+            # Unintuitevly a missing template will throw a
+            # UnboundLocalError as the missing file will kick the
+            # migration routine for zfs props. We don't need that :)
+            raise RuntimeError(f"Template: {self.release} not found!")
+
+        return config_fn
+
+    def _get_clone_config_fn(self, jail_uuid, path):
+        config_fn = iocage.lib.ioc_json.IOCJson(path).json_get_value
+        try:
+            config_fn("cloned_release")
+        except (IOError, OSError, FileNotFoundError, UnboundLocalError):
+            if os.path.isdir(f"{self.iocroot}/templates/{self.release}"):
+                iocage.lib.ioc_common.logit({
+                    "level"  : "EXCEPTION",
+                    "message": "You cannot clone a template, "
+                               "use create -t instead."
+                },
+                    _callback=self.callback,
+                    silent=self.silent)
+            else:
+                # Yep, self.release is actually the source jail.
+                iocage.lib.ioc_common.logit({
+                    "level"  : "EXCEPTION",
+                    "message": f"Jail: {self.release} not found!"
+                },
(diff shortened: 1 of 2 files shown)


iocage/lib/ioc_create.py
```diff
@@ -86,6 +86,106 @@ def create_jail(self):
             iocage.lib.ioc_destroy.IOCDestroy().destroy_jail(location)
             sys.exit(1)
 
+    def _get_template_config_fn(self, jail_uuid, path):
```
igalic, line 89: from what i gather from other code, this function name should be `__get_template_config_fn__`
although, i'd much rather it be `__get_config_fn__` ;)

iocage/lib/ioc_create.py
```diff
@@ -86,6 +86,106 @@ def create_jail(self):
...
+
+        except (IOError, OSError, FileNotFoundError, UnboundLocalError):
+            iocage.lib.ioc_common.logit({
+                "level"  : "EXCEPTION",
+                "message": f"RELEASE: {self.release} not found!"
+            },
+                _callback=self.callback,
+                silent=self.silent)
+
+    def _su_check_call(self, jail_uuid, path, _type):
```
igalic, line 150: this is a really bad name, but i didn't know what to call it at the time.

iocage/lib/ioc_create.py
```diff
@@ -86,6 +86,106 @@ def create_jail(self):
...
+                # Yep, self.release is actually the source jail.
+                iocage.lib.ioc_common.logit({
+                    "level"  : "EXCEPTION",
+                    "message": f"Jail: {self.release} not found!"
+                },
+                    _callback=self.callback,
+                    silent=self.silent)
+        return config_fn
+
+    def _get_release_config(self, jail_uuid, path):
```
igalic, line 124: this one's tricky, as it's very different from all the other ones.

i'm not calling it a smell, just, like, production code that deals with realities ;)

iocage/lib/ioc_create.py
```diff
@@ -86,6 +86,106 @@ def create_jail(self):
...
+                },
+                    _callback=self.callback,
+                    silent=self.silent)
+
+            except libzfs.ZFSException:
+                pass
+            raise RuntimeError(err)
+
+    def _zfs_clone(src, dst, *opts):
+        su.Popen(["zfs", "clone", opts, src, dst],
```
igalic, line 186: puppet has a really nice concept of creating "commands" like so:

```ruby
  commands zfs: "zfs"
```
which basically wraps the 80% case of just calling `exec('zfs', params)`
but… more important might be wether `*opts` is the way to go for python ;)

igalic (commented):
a lot of naming issues in this code…

skarekrow:
I won't be able to review this for a little bit, quite busy outside of iocage. Thanks for the PR! I'll try to get to it soon 👍

igalic:
i'll have some more, i hope i don't overwhelm you too much
ioc_fetch and other systems will need a similar treatment

skarekrow:
It's quite nice to see PR's like this. I haven't had the time to tighten it up myself, so any help is appreciated. 

Though I will say I would hold off if I were you. The intention of @gronke's PR is to replace everything in the `lib/` directory eventually. 

So unless it really bugs you, it may be a lot of work that's going to be deprecated sooner than later. But feel free to continue if you want!

igalic:
my intent was primarily to fix the #290 but when trying to do that, i had serious trouble understanding the underlying code, and hence these prs.

i didn't realize that @gronke's work was going in the same direction as mine. in that sense, i should probably help with #267, rather than continuing this.

igalic:
closing in favour of #267

PR #336, opened 2017-08-20
Only specify ip[46].addr on jail command line if set

jhujhiti:
Previously we would send `jail -c ip4.addr=`, which disables ip4, breaking iocage's inherit option. Inherit works if ip4.addr is completely unspecified on the jail command line.


skarekrow:
Thanks!

Issue #389, opened 2017-10-14
man page: clone doesn't explain [-c | --count]


wxcafe:
- [ ] Supply `iocage --version` N/A
- [ ] Supply the commands used, along with any steps to recreate it. N/A
- [ ] Provide the output from the command you issued. N/A
- [ ] Supply what you expected the result or output to be N/A
- [x] Checked that the problem has not already been fixed on `master` if using
 a stable release.

In the iocage manpage (from September 5, 2017), at the top (subcommands listing) it's specified that clone can take a [-c | --count] argument, but in the detailed description of clone (in the Subcommands section) there is no explanation for this. The explanation is only in create (I found this while in the process of writing ZSH completion files for #160)

PR #395, opened 2017-10-16
Adds [-c | --count] flag documentation to clone subcommand

wxcafe:
- [x] Explain the feature
    documents [-c | --count] flag for the clone subcommand as explained in iocage/iocage#389


skarekrow:
Thanks!

wxcafe:
Sure! btw, I asked in the IRC channel but got no response: are the _all, _pool, _long, and so on in the various "options" fields on purpose? like this : 
screenshot from 2017-10-16 22-31-25

skarekrow:
I’d ignore those. They’re internal as they would collide in namespaces. Just stick to the short and long flags.

wxcafe:
I mean, yeah, but I'm pretty sure they should be removed from the man page, since they're not in the --help output and can't be used

skarekrow:
Hah yes absolutely. @Mrt134 ping :D

PR #436, opened 2017-12-03
Misc improvements for building from git

asomers:
Several improvements to the build process when building from git:

* Allow source tree to be located anywhere
* Don't install dependencies in the Makefile
* Update the submodule, if it isn't already up-to-date
* Update the py-libzfs submodule


skarekrow (approved):

skarekrow:
Thanks!

PR #455, opened 2017-12-31
Don't assume the sys.prefix is /usr/local

jhujhiti:
I install my ports into /usr/pkg. The iocage port ended up installing zsh completion, the rc script, and the man page to /usr/local because it was hardcoded in setup.py. This PR fixes that.

Setuptools assumes that relative paths in `data_files` belong under `sys.prefix`, which is exactly what we want here, so we can simply remove the leading `/usr/local/` and adjust the logic to check for the existence of the correct directories.


skarekrow (approved):

skarekrow:
Thanks!

PR #471, opened 2018-01-17
More helpful message when snapshot already exists

aduitsis:
When trying to` iocage -f list` to transition from an old non-python iocage installation, if the snapshot that is being created as part of the upgrade process already exists, a generic error will be printed and there is no way (that I know of) to get out of that situation by using iocage alone. By providing the exact snapshot that already exists, the user can e.g. choose to `zfs rename -r` said snapshot and thus overcome the impasse.

### DIFF
--- a/iocage/lib/ioc_json.py
+++ b/iocage/lib/ioc_json.py
@@ -1447,8 +1447,9 @@ class IOCJson(object):
                         if err.code == libzfs.Error.EXISTS:
                             iocage.lib.ioc_common.logit(
                                 {
-                                    "level": "EXCEPTION",
-                                    "message": 
+                                    "level": 
+                                    "EXCEPTION",
+                                    "message":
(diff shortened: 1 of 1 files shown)


skarekrow:
Thanks for the PR! If you could break the line up, I'll be able to merge this as it's too long :)

skarekrow:
Apparently Travis is still unhappy with the line length :D 

Third times the charm ;)

aduitsis:
Line 1492 does not have a trailing whitespace anymore, any idea why the check insists it does?

skarekrow:
Don't worry about that, I'll actually be grabbing this and rebasing it locally. I just noticed this was on the wrong branch ;)

Thanks again for the patch!

Merged here:https://github.com/iocage/iocage/commit/baebfbfcc4e86efcb3ca80300073e3c478d9269d

PR #485, opened 2018-02-13
Do not falsely report zfs exceptions when creating template

worr:
Currently, when converting a jail into a template, if there are any errors
setting zfs options or renaming datasets, iocage will report that a template
of the same name exists.

This changes the behavior to throw an exception when we encounter this case,
like the opposite case (template -> jail) does.

- [x] Explain the feature

### DIFF
--- a/iocage/lib/ioc_json.py
+++ b/iocage/lib/ioc_json.py
@@ -651,64 +651,64 @@ class IOCJson(object):
 
                 if value == "yes":
                     try:
-                        try:
-                            jail_zfs_dataset = f"{pool}/" \
-                                f"{conf['jail_zfs_dataset']}"
-                            self.zfs_set_property(jail_zfs_dataset, "jailed",
-                                                  "off")
-                        except libzfs.ZFSException as err:
-                            # The dataset doesn't exist, that's OK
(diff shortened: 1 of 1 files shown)


iocage/lib/ioc_json.py
```diff
@@ -651,64 +651,50 @@ def json_set_value(self, prop, _import=False, default=False):
...
-                                "level": "INFO",
-                                "message": f"{uuid} converted to a template."
-                            },
-                            _callback=self.callback,
-                            silent=self.silent)
+                        if err.code == libzfs.Error.NOENT:
+                            pass
+                        else:
+                            # Danger, Will Robinson!
+                            raise
```
skarekrow: Since we're handling other exceptions, let's do this here:
```
iocage.lib.ioc_common.logit(
  {
    "level": "EXCEPTION",
    "message": err
  },
  _callback=self.callback)
```
So the user can get a nicely formatted error, and the lib correctly passes the exception.

skarekrow (changes requested):

worr:
Gotcha, I've added some more exception handling in the latest diff.

skarekrow (approved):

skarekrow:
Thanks!

PR #504, opened 2018-03-09
Remove DS_Store files

0mp:
None of those files looks like it is  super important ;)


skarekrow:
Haha indeed. Wonder how those got in there.

skarekrow (approved):

Issue #512, opened 2018-03-24
empty jail creation gives traceback error


Zenzou:
I'm currently on Freenas 11.1-U4 and trying out iocage.
I've successfully created jails using --plugins (deluge,emby), and created normal and basejails (unifi5 controller)
I tried to create a debian jail,
`http://iocage.readthedocs.io/en/latest/debian.html`
however 'iocage create -e' gives a traceback.
```
Traceback (most recent call last):
  File "/usr/local/bin/iocage", line 10, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/iocage/cli/create.py", line 103, in cli
    elif release.lower() == "latest":
AttributeError: 'NoneType' object has no attribute 'lower'
```
Looks like they're using iocage 1.0 ALPHA

jsegaert:
This particular traceback was already reported and fixed in #507.  
However, creating an empty jail using the latest version of iocage also seems to be an issue, but it produces a different traceback:
```
/root root@trinity#  iocage create -e -n debjail 
Traceback (most recent call last):
  File "/usr/local/bin/iocage", line 10, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #531, opened 2018-04-26
Moved up logic for "empty jail" to fix issue 512

jsegaert:

- [X] Explain the feature
I moved up the logic/check for "empty jails" to fix issue #512
It will still create an fstab file, but not a hosts file.  
After the fix, I managed to create Debian Wheezy jail as per the original issue report.


skarekrow (approved):
Thanks!

PR #538, opened 2018-05-02
add "hosid_strickt_check" as a new value

ole-db:
This PR adds "hosid_strickt_check" as a new value.
The default is set to "off".
If the value is set set to "on" the jail will only start if `hostid` matches `/etc/hostid`

This was dicussed in https://github.com/iocage/iocage/issues/529


skarekrow (approved):

skarekrow:
Looks great, thanks!

PR #541, opened 2018-05-04
Changed chroot command for better security

Avatat:
Previous `iocage chroot` command use `chroot` command, but it gives only filesystem level isolation. When some malicious binary from jail root will be runned, it can jump out of chroot space and do something above jail process space.
Better option is usage of `jexec` to run shell inside jail root.

### DIFF
--- a/iocage/lib/iocage.py
+++ b/iocage/lib/iocage.py
@@ -422,7 +422,10 @@ class IOCage(object):
                 _callback=self.callback,
                 silent=self.silent)
 
-        chroot = su.Popen(["chroot", f"{path}/root"] + command)
+        if len(command) == 0:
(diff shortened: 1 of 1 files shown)


gronke:
This is pretty much what `iocage console` and `iocage exec` are already doing. I'd favor to deprecate the whole function and remove the CLI command. I agree that chrooting into a jail dataset is a security risk and in my opinion obsoletes the purpose of jails.

skarekrow:
I agree, it's better to change the one usage in the program for `upgrade`, though I may be forgetting other usages OTOH.

gronke:
@Avatat I'll close this PR for now. You have my full support for another one removing chroot entirely. Whatever @skarekrow could have overseen, it can't be worth the risk that comes with chrooting into an untrusted environment.

PR #543, opened 2018-05-05
execute pip via python3.6 -m pip in Makefile

gronke:
Depending on the install method of pip the name of the executable differs. The safest way is to only rely on the python executables name while loading pip via `-m` flag. For example:

```
python3.6 -m pip install -U .
```


skarekrow (approved):

skarekrow:
Thanks!

Issue #468, opened 2018-01-14
iocage doesn't rename fstab when renaming jail


dapperfu:
Make sure to follow and check these boxes before submitting an issue! Thank you.

- [X] Supply `iocage --version`

    Version	0.9.10 2017/12/22

- [X] Supply the commands used, along with any steps to recreate it.

```

root@functionalsafety:~/FreeNAS-devops # iocage rename 3acf803d nginx-lite
Jail: 3acf803d renamed to nginx-lite
root@functionalsafety:~/FreeNAS-devops # iocage start nginx-lite
```

- [X ] Provide the output from the command you issued.

```
* Starting nginx-lite-f3d1-48ad-b4d8-571455cd0ac2
  + Start FAILED
mount: /iocage/jails/3acf803d-f3d1-48ad-b4d8-571455cd0ac2: No such file or directory
jail: /sbin/mount -t nullfs -o ro /iocage/releases/11.1-RELEASE/root/bin /iocage/jails/3acf803d-f3d1-48ad-b4d8-571455cd0ac2/root/bin: failed

```

- [X] Supply what you expected the result or output to be

The renamed jail starts correctly.

- [ ] Checked that the problem has not already been fixed on `master` if using
 a stable release.

skarekrow:
Closed by #547 thanks to @jsegaert <3

PR #547, opened 2018-05-09
Make renaming jails more robust

jsegaert:

- [X] Explain the feature
- Will adjust the mountpoints in fstab when renaming a jail. Please refer to issues #468 and #525 
- Fixes the mountpount when renaming jail using partial jailname (i.e. when only using first characters of uuid)


skarekrow (approved):
Looks good, thanks!

Issue #532, opened 2018-04-26
Feature Suggestion: Force stopping a jail


vendion:
If would be nice to be able to force stop a jail with `iocage` rather than having to fall back to `jail -r <JID>` to do so. Some reasons why this may be helpful could be the jail stops responding or something goes majorly wrong and you need to rollback no a snapshot.

A very terrible example might by
```
$ sudo iocage create -n test
$ sudo iocage start test
$ sudo iocage snapshot test
$ sudo iocage console test
# rm -rf /* // inside the test jail
(new terminal/window/whatever)
$ sudo iocage stop test // rollback requires jail to be stopped
* Stopping test
  + Running prestop OK
Traceback (most recent call last):
  File "/usr/local/bin/iocage", line 10, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/iocage/cli/stop.py", line 54, in cli
    ioc.IOCage(exit_on_error=True, jail=jail, rc=rc).stop()
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/iocage.py", line 1655, in stop
    exit_on_error=self.exit_on_error)
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/ioc_stop.py", line 52, in __init__
    self.__stop_jail__()
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/ioc_stop.py", line 142, in __stop_jail__
    stdout=f, stderr=su.PIPE)
  File "/usr/local/lib/python3.6/subprocess.py", line 291, in check_call
    raise CalledProcessError(retcode, cmd)
[...]

PR #552, opened 2018-05-12
Added forced stop

Avatat:
Related to: https://github.com/iocage/iocage/issues/532
I added force option to `iocage stop` command. It skip all pre-stop actions, kill jail process and do post-stop actions.

I don't know why git recognize so many lines as changed in `iocage/lib/ioc_stop.py`, because I added only few lines and indented many other lines.. I will be glad if someone explain me this - I'm new in git world :)
EDIT: Probably it is related to tab and space indentions.. How can I fix it? @skarekrow could you help me?

### DIFF
--- a/iocage/cli/stop.py
+++ b/iocage/cli/stop.py
@@ -34,10 +34,8 @@ __rootcmd__ = True
 @click.option("--rc", default=False, is_flag=True,
               help="Will stop all jails with boot=on, in the specified"
                    " order with higher value for priority stopping first.")
-@click.option("-f", "--force", default=False, is_flag=True,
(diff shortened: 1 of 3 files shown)


skarekrow (changes requested):
Looks good but a couple changes:
1 - force should be a kwarg and placed after the rest to minimize disturbance with existing API usage
2 - We should elaborate on the force help text describing what it's doing.

Avatat:
@skarekrow, I moved my changes to new branch (from my master branch) and created new PR: https://github.com/iocage/iocage/pull/553

I changed type of `force` arguments and added better help text.

Issue #533, opened 2018-04-29
Unable to rename templates.


pjd:
Version 0.9.10 2017/12/22
beast:root:~# iocage rename builder-freebsd10 builder-freebsd10.4
Traceback (most recent call last):
  File "/usr/local/bin/iocage", line 10, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/iocage/cli/rename.py", line 39, in cli
    ioc.IOCage(exit_on_error=True, jail=jail).rename(new_name)
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/iocage.py", line 1211, in rename
    self.set(f"host_hostuuid={new_name}", rename=True)
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/iocage.py", line 1418, in set
    iocjson.json_set_value(prop)
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/ioc_json.py", line 723, in json_set_value
    self.json_write(conf)
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/ioc_json.py", line 413, in json_write
    'w') as out:
  File "/usr/local/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/ioc_common.py", line 420, in open_atomic
    with tempfile(dir=os.path.dirname(os.path.abspath(filepath))) as tmppath:
  File "/usr/local/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/ioc_common.py", line 383, in tempfile
[...]

PR #570, opened 2018-06-15
Rename templates

jsegaert:

- [X] Explain the feature
This PR adds support for renaming templates, as per #533   The main trick was to temporary flip the "readonly" zfs property while renaming

Full disclose with additional fixes/features:
- fix typo "runnning" should be running"
- added validation to ensure that you cannot use a name that is already taken, i.e. name that already exists as jail or template (identical to the check that happens during "create")
- I found the "path.replace(uuid, new_name)" to be problematic when using generic names such as "jail" of "template" or any name that (even partially) matched a substring in the mountpoint directory; That was made that more robust
- it had hardcoded to only handle the default jail_zfs_dataset {jail_path}\data and this was beefed up to actually properly parse property **jail_zfs_dataset** to process the actual zfs_datasets
- will always show the full jail name in the message "Jail: {} rename to {}", even when a partial name was used.
- The property **host_hostuuid** will now be changed only _after_ the zfs rename is successful, to prevent that the jail ends in some insonsistent state (there are some reports of that in a few of the rename-related issues)
- if the hostname (i.e. property **host_hostname**) was identical to the the old jail name, then host_hostname will also be changed to be the new name.  There's similar logic in some of the migration methods and it prevent things from getting confusing when running the native jls command  

### DIFF
--- a/iocage/lib/iocage.py
+++ b/iocage/lib/iocage.py
@@ -1156,7 +1156,7 @@ class IOCage(object):
         new_mountpoint = f"{self.iocroot}/{_folders[0]}/{new_name}"
 
         if (os.path.isdir(new_mountpoint) or
-            os.path.isdir(f"{self.iocroot}/{_folders[1]}/{new_name}")):
+                os.path.isdir(f"{self.iocroot}/{_folders[1]}/{new_name}")):
 
             ioc_common.logit(
                 {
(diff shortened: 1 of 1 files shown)


skarekrow (approved):
Thanks for all the PR's :)

PR #574, opened 2018-06-24
Fix validation of 'set template'

### LINKED ISSUE #573: Incorrect behavior of 'set template'

Latest master

- [X] Supply the commands used, along with any steps to recreate it.

```
/root root@trinity#  iocage create -n foo -r latest
foo successfully created!

/root root@trinity#  iocage set template=true foo
foo is already a jail!

/root root@trinity#  iocage set template=on foo
foo is already a jail!

/root root@trinity#  iocage set template=yes foo
foo converted to a template.
Property: template has been updated to yes

/root root@trinity#  iocage set template=false foo
foo is already a template!
```
- [X] Provide the output from the command you issued.
When providing an invalid value for property "template", it responds with "XXX is already a jail"

- [X] Supply what you expected the result or output to be
I expected it to respond "Value must be yes or no", similar to what happens when attempting to update other properties, e.g.:
```
/root root@trinity#  iocage set boot=false foo
false is not a valid value for boot.
Value must be off or on
```
- [X] Checked that the problem has not already been fixed on `master` if using
 a stable release.

jsegaert:

- [X] Explain the feature

This is to resolve #573 

### DIFF
--- a/iocage/lib/iocage.py
+++ b/iocage/lib/iocage.py
@@ -1399,7 +1399,9 @@ class IOCage(object):
             return
 
         if "template" in key:
-            if "templates/" in path and prop != "template=no":
+            if prop == "template=yes" and path.startswith(
+                    f"{self.iocroot}/templates/"):
+
                 ioc_common.logit(
                     {
                         "level": "EXCEPTION",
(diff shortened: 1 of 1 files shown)


skarekrow (approved):
Thanks!

Issue #452, opened 2017-12-20
Can't modify template after creation


tehfink:
Is conversion to a template now a one-way street?

```
# iocage --version
Version 0.9.9.2 09/30/2017
# iocage create --name template-02 --pkglist=/iocage/pkgs.json -r 11.1-RELEASE boot=off allow_sysvipc=1 allow_raw_sockets=1 host_hostname=template-02 template=yes ip4_addr="lo1|172.16.10.1/28"
template-02 successfully created!
Testing SRV response to FreeBSD
Testing DNSSEC response to FreeBSD

Installing pkg... 
Installing supplied packages:
  - nano... 
  - screen... 
Message from screen-4.6.1_2:
…snipped
===========================================================================
# iocage update template-02
Please convert back to a jail before trying to update template-02
# iocage set template=no template-02
Traceback (most recent call last):
  File "/usr/local/bin/iocage", line 10, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/iocage/cli/set.py", line 45, in cli
    plugin)
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/iocage.py", line 1359, in set
    iocjson.json_set_value(prop)
  File "/usr/local/lib/python3.6/site-packages/iocage/lib/ioc_json.py", line 599, in json_set_value
    self.zfs.get_dataset(new_location).rename(old_location)
  File "libzfs.pyx", line 2106, in libzfs.ZFSObject.rename
libzfs.ZFSException: Device busy
# iocage update template-02
Please convert back to a jail before trying to update template-02
# iocage stop template-02
template-02 is not running!
# iocage list -t
+-----+-------------+-------+--------------+-------------+
| JID |    NAME     | STATE |   RELEASE    |     IP4     |
[...]

PR #590, opened 2018-08-02
Always force unmount ZFS-datasets when renaming

jsegaert:

- [X] Explain the feature

- This fixes issue #452 by setting the **forceunmount** flag when renaming the ZFS-dataset during conversion between jail and template.  
- There are two other calls to that same rename method that got the same treatment.
- Fix typo: "opertaions" should be "operations"


skarekrow (approved):

skarekrow:
Great! Thanks :)

PR #591, opened 2018-08-04
Dynamic jail config merging

gronke:
- When a jail is not configured with a property, the default value is used
- Refactor ioc_json
- Refactor iocroot handling
- Utilize py-libzfs more often

This changes ensure compatibility with libiocage jails where only configuration changes by the user get stored in `config.json`. By falling back to the hosts (or hardcoded) defaults, it is no longer required to store all config properties in a jail.


skarekrow:
We should target this at the v2 branch, that way any regressions can be caught there.

gronke:
This work is not affecting v2 using libiocage. But it makes v1 compatible with partial `config.json` files, so that it is possible to start jails, that were created with libiocage, with v1 of python-iocage.

gronke:
Hi there, @skarekrow! How can I make it easier for you to review the changes? I volunteer to setup a test-machine for us where we can share a tmux and walk through... almost everything. Sounds good?

iocage_lib/ioc_create.py
```diff
@@ -326,17 +350,20 @@ def _create_jail(self, jail_uuid, location):
...
-            ).__destroy_parse_datasets__(
-                f"{self.pool}/iocage/jails/{jail_uuid}")
-            iocage_lib.ioc_common.logit({
-                "level": "EXCEPTION",
-                "message": f"You cannot name a jail {jail_uuid}, "
-                           "that is a reserved name."
-            },
-                _callback=self.callback,
-                silent=self.silent)
+        if not self.plugin:
```
skarekrow, line 353: Why this check? The jail names are reserved for plugins or jails, that doesn't matter.

iocage_lib/ioc_create.py
```diff
@@ -359,37 +389,38 @@ def _create_jail(self, jail_uuid, location):
...
-                iocjson.json_write(config)  # Destroy counts on this.
-                iocage_lib.ioc_destroy.IOCDestroy().destroy_jail(location)
-
-                raise RuntimeError(f"***\n{err}\n***\n")
-            except SystemExit:
-                iocjson.json_write(config)  # Destroy counts on this.
-                iocage_lib.ioc_destroy.IOCDestroy().destroy_jail(location)
-                exit(1)
+                try:
+                    iocjson.json_check_prop(key, value, config)
```
skarekrow, line 393: Why the removal? We want value and config returned here. For instance `ip4_addr` will possibly be `DEFAULT` and the user will want a default gateway set.

iocage_lib/ioc_fetch.py
```diff
@@ -264,19 +262,23 @@ def fetch_release(self, _list=False):
...
-            dataset = f"{self.iocroot}/download/{self.release}"
-            pool_dataset = f"{self.pool}/iocage/download/{self.release}"
+            pool_dataset_name = f"{self.pool}/iocage/download/{self.release}"
 
-            if os.path.isdir(dataset):
+            try:
+                dataset = self.zfs.get_dataset(
+                    f"{self.iocroot}/download/{self.release}"
+                )
+                return
```
skarekrow, line 271: This changes the behavior of fetch. We check if it's a directory to see if it's mounted (that can be improved), but returning here means the fetch process will not re-extract and update the RELEASE.

iocage_lib/ioc_fetch.py
```diff
@@ -341,6 +343,7 @@ def fetch_http_release(self, eol, _list=False):
             self.server = "http://" + self.server
 
         logging.getLogger("requests").setLevel(logging.WARNING)
+        print("fetch_http_release")
```
skarekrow, line 346: Debug

iocage_lib/ioc_json.py
```diff
@@ -424,85 +455,98 @@ def _upgrade_pool(self, pool):
...
         if prop == "pool":
-            match = 0
-
-            for pool in zpools:
+            raise DeprecationWarning("Access ioc_root.pool directly")
+        elif prop == "iocroot":
+            raise DeprecationWarning("Access ioc_root.iocroot directly")
+            return self.iocroot.mountpoint
+        elif prop == "all":
+            raise DeprecationWarning("Access the IOCJson instance directly")
```
skarekrow, line 466: Why the deprecation warnings? Hardly a warning when it's fatal :P Also this interface is not deprecated in iocage, and this will break the existing FreeNAS middleware plugin.

iocage_lib/ioc_json.py
```diff
@@ -543,304 +587,314 @@ def json_get_value(self, prop, default=False):
                     iocage_lib.ioc_common.logit(
                         {
                             "level":
-                            "INFO",
+                            "EXCEPTION",
                             "message":
-                            f"Setting up zpool [{zpool}] for"
-                            " iocage usage\n If you wish to change"
-                            " please use \"iocage activate\""
+                            "IOCAGE_SKIP is TRUE or an RC operation, not"
```
skarekrow, line 592: We're duplicating this EXCEPTION?

iocage_lib/ioc_json.py
```diff
@@ -543,304 +587,314 @@ def json_get_value(self, prop, default=False):
...
-            iocage_lib.ioc_common.logit(
-                {
-                    "level": "INFO",
-                    "message": f"Property: {key} has been updated to {value}"
-                },
-                _callback=self.callback,
-                silent=self.silent)
-
-            # We can attempt to set a property in realtime to jail.
+                    self.lgr.disabled = True
```
skarekrow, line 786: Why are we disabling the logger? In iocage that's most cli output.

iocage_lib/ioc_start.py
```diff
@@ -857,7 +892,7 @@ def start_copy_localtime(self):
 
     def start_generate_resolv(self):
         resolver = self.get("resolver")
-        #                                     compat
+        # compat
```
skarekrow, line 895: That whitespace was meant to directly reference the `resolver != none` bit below it :) So it served a purpose.

skarekrow (changes requested):

skarekrow:
This is now done in https://github.com/iocage/iocage/commit/8ad2703542bab9364be8f157d30bcb683b64842e

PR #603, opened 2018-08-28
fixes BUG, return was one level high

ole-db:
fixes BUG, return was one level high. Otherwise Jails wont start.


skarekrow (approved):
Thanks! Good catch

PR #607, opened 2018-09-06
Bug fix for installing packages

sonicaj:
This commit fixes a bug which caused an exception when trying to install plugins or packages with no packages specified to be installed.
Ticket: #44735

### DIFF
--- a/iocage_lib/ioc_create.py
+++ b/iocage_lib/ioc_create.py
@@ -648,6 +648,8 @@ class IOCCreate(object):
             _callback=self.callback,
             silent=supply_msg[1])
 
+        pkg_err_list = []
+
         for pkg in self.pkglist:
             iocage_lib.ioc_common.logit({
                 "level": "INFO",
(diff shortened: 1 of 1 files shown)


gronke (changes requested):
The function should terminate with an Exception rather than returning a concatenated string of error messages on failure.

Instead of handling a list of errors we can just bundle the package list into a single `pkg install`, so that only a single error can occur when installing packages? If an error occurs, I would expect an Exception being raised immediately.

sonicaj:
I agree to this. I can make the necessary changes but not sure if it is necessary to do the os.remove which is being done after the loop and if the behavior of the exceptions needs to be changed as well. In case of plugins, we don't raise one, we return the error message and else we raise one. I have looked at usages and we can change that but I am not sure of the motivation behind this design so didn't want to take such a step. If you could kindly fill me in on that, I can update the PR accordingly

skarekrow (approved):

skarekrow:
Merging as freenas has this already, that's too breaking of a change currently.

PR #621, opened 2018-09-25
Fixed iocage list multiple IP4 addresses

thedix:
This pull request resolves iocage list problem when multiple IP4 addresses specified for jail.
For example, setting ip4_addr="lo1|192.168.1.1,lo1|192.168.2.2" leads to showing "192.168.1.1,lo1" instead of correct "192.168.1.1,192.168.2.2".

### DIFF
--- a/iocage_lib/ioc_list.py
+++ b/iocage_lib/ioc_list.py
@@ -170,7 +170,7 @@ class IOCList(object):
             ip6 = conf["ip6_addr"]
 
             try:
-                short_ip4 = full_ip4.split("|")[1].split("/")[0]
+                short_ip4 = ",".join([item.split("|")[1].split("/")[0] for item in full_ip4.split(",")])
             except IndexError:
                 short_ip4 = full_ip4 if full_ip4 != "none" else "-"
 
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_list.py
```diff
@@ -170,7 +170,7 @@ def list_all(self, jails):
             ip6 = conf["ip6_addr"]
 
             try:
-                short_ip4 = full_ip4.split("|")[1].split("/")[0]
+                short_ip4 = ",".join([item.split("|")[1].split("/")[0] for item in full_ip4.split(",")])
```
skarekrow: Longer then 79 characters, do a line break on the `"/"` split like this:
```
                short_ip4 = ",".join([item.split("|")[1].split(
                    "/")[0] for item in full_ip4.split(",")])
```

skarekrow (changes requested):

thedix:
Fixed. Is it ok now?

skarekrow (approved):

skarekrow:
Thanks!

Issue #608, opened 2018-09-08
Jail fails to start when depends jail is already running


davefullard:
Make sure to follow and check these boxes before submitting an issue! Thank you.

- [X] Supply `iocage --version`
- [X] Supply the commands used, along with any steps to recreate it.
- [X] Provide the output from the command you issued.
- [X] Supply what you expected the result or output to be
- [X] Checked that the problem has not already been fixed on `master` if using
 a stable release.

### Version
Version 0.9.10 2017/12/22

### Commands used and output
```
[dave@tardis ~]$ sudo iocage set depends="pgsql" nextcloud
Property: depends has been updated to pgsql
[dave@tardis ~]$ sudo iocage start nextcloud
pgsql is already running!
[dave@tardis ~]$
```

### Expected output
```
[dave@tardis ~]$ sudo iocage start nextcloud
* Starting nextcloud
  + Started OK
  + Starting services OK
```

No previous bug has been filed.

skarekrow:
Closed by #622

PR #622, opened 2018-09-25
 Fixed bug: Jail fails to start when depends jail is already running (#608)

thedix:
The idea is to try starting all the nested jails.

### DIFF
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -42,6 +42,7 @@ def callback(_log, callback_exception):
     log = logging.getLogger("iocage")
     level = _log["level"]
     message = _log["message"]
+    force_raise = _log.get("force_raise")
 
     if level == 'CRITICAL':
         log.critical(message)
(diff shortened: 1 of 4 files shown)


skarekrow (changes requested):
We should create a new exception and return that when a jail is already running. Then we can reliably catch it here. See ioc_exceptions for that

thedix:
Ok, I have to take some time for this.

thedix:
Check this out.

I added new exception JailRunning and new property IOCStart.is_depend which is True for all the nested jail when starting. If is_depend==True then IOCStart.__start_jail__() raises JailRunning for starting jail. If is_depend==False then __start_jail__() raises SystemExit(1) as usual (this is a parent jail). Also I modified default callback() function and added new argument "force_raise" to force raising exception when there is a valid TTY. Previous code raised exception only without TTY. Method IOCage.start() processes all the dependent jails with setting is_depend=True for each and reverting previous value after call finished. 

Now the logic is the following: parent jail is not started if depended jails failed to start. If depended jails are already started then start parent jail.

skarekrow (approved):

skarekrow:
Great change, thanks!

PR #624, opened 2018-09-27
Fixed bug: setting ZFS properties via 'iocage set' raised exception

thedix:
Setting ZFS properties via 'iocage set' leads to raising exception because of lacking return statement at json_check_prop().

```
# iocage set quota=8g test

...
  File "/usr/local/lib/python3.6/site-packages/iocage_lib/ioc_json.py", line 759, in json_set_value
    value, conf = self.json_check_prop(key, value, conf)
TypeError: 'NoneType' object is not iterable
```


skarekrow (approved):

skarekrow:
Good catch! Thanks

PR #633, opened 2018-10-05
libiocage: activate and deactivate command

gronke:
- Use libiocage in `activate` and `deactivate` commands

### DIFF
--- a/iocage_cli/activate.py
+++ b/iocage_cli/activate.py
@@ -46,13 +46,9 @@ def cli(ctx, zpool, mountpoint):
         logger.error("Cannot activate when executed with explicit sources.")
         exit(1)
 
-    iocage_pool = None
-
-    for pool in zfs.pools:
-        if pool.name == zpool:
-            iocage_pool = pool
-
-    if iocage_pool is None:
+    try:
+        iocage_pool = zfs.get(zpool)
+    except Exception:
         logger.error(f"ZFS pool '{zpool}' not found")
         exit(1)
 
--- a/iocage_cli/deactivate.py
+++ b/iocage_cli/deactivate.py
@@ -41,13 +41,9 @@ def cli(ctx, zpool):
     logger = ctx.parent.logger
     zfs = ctx.parent.zfs
 
-    iocage_pool = None
-
-    for pool in zfs.pools:
-        if pool.name == zpool:
-            iocage_pool = pool
-
-    if iocage_pool is None:
+    try:
+        iocage_pool = zfs.get(zpool)
+    except Exception:
         logger.error(f"ZFS pool '{zpool}' not found")
         exit(1)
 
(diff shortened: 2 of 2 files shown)


iocage_cli/activate.py
```diff
@@ -21,22 +22,49 @@
...
+
+    iocage_pool = None
+
+    for pool in zfs.pools:
+        if pool.name == zpool:
+            iocage_pool = pool
+
+    if iocage_pool is None:
+        logger.error(f"ZFS pool '{zpool}' not found")
+        exit(1)
```
william-gr, line 53: We could do:

```
try:
    iocage_pool = zfs.get(zpool)
except Exception:
    logger.error(f"ZFS pool '{zpool}' not found")
    exit(1)
```

iocage_cli/activate.py
```diff
@@ -21,22 +22,49 @@
...
+        datasets = iocage.Datasets.Datasets(
+            zfs=zfs,
+            logger=logger
+        )
+        datasets.activate_pool(
+            pool=iocage_pool,
+            mountpoint=mountpoint
+        )
+        logger.log(f"ZFS pool '{zpool}' activated")
+    except iocage.errors.IocageException:
```
william-gr, line 65: Shouldnt we print something to the user in case that happens? Does that exception have an error message?

gronke: The default logger always prints `error` and `warn` levels, so that each such error will already have printed to the users screen using the CLI. 

When an `iocage.error.IocageException` occurs, we know that the user was notified already.

iocage_cli/activate.py
```diff
@@ -21,22 +22,45 @@
...
 import click
 
-import iocage_lib.ioc_common as ioc_common
-import iocage_lib.iocage as ioc
+import iocage.errors
+import iocage.Datasets
+import iocage.Logger
+import iocage.ZFS
 
 __rootcmd__ = True
```
igalic, line 35: how come pep8's not complaining about so many spaces here 😁

iocage_cli/deactivate.py
```diff
@@ -0,0 +1,62 @@
...
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+# IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+"""Deactivate ZFS pools for iocage with the CLI."""
```
igalic, line 25: i didn't even know we had that!

igalic (approved):
👀

skarekrow (approved):

iocage_cli/activate.py
gronke: It would in case you choose any other number of newlines.

PR #634, opened 2018-10-05
libiocage: snapshot command

gronke:
- Transfer the snapshot command from libiocage/ioc


gronke:
Wrong base branch `master`, should have been `v2`.

PR #636, opened 2018-10-05
libiocage: get and set commands

gronke:
- copy get and set commands from libiocage/ioc

### DIFF
--- a/iocage_cli/get.py
+++ b/iocage_cli/get.py
@@ -47,16 +47,10 @@ from .shared.click import IocageClickContext
     help="Get all properties for the specified jail.",
     is_flag=True
 )
-@click.option(
-    "--pool", "-p", "_pool",
-    help="Get the currently activated zpool.",
-    is_flag=True
-)
 def cli(
     ctx: IocageClickContext,
     prop: typing.Tuple[str],
     _all: bool,
-    _pool: bool,
     jail: typing.Optional[str]
 ) -> None:
     """Get a list of jails and print the property."""
@@ -64,14 +58,6 @@ def cli(
     host = iocage.Host.Host(logger=logger)
 
     _prop = None if len(prop) == 0 else prop[0]
-
-    if _pool is True:
-        try:
-            print(host.datasets.active_pool.name)
-        except Exception:
-            print("No active pool found")
-            exit(1)
-        exit(0)
 
     if _all is True:
         if jail is None:
--- a/iocage_cli/get.py
+++ b/iocage_cli/get.py
@@ -37,7 +37,11 @@ from .shared.click import IocageClickContext
 @click.command(
     context_settings=dict(max_content_width=400,),
     name="get",
-    help="Gets the specified property."
+    help="""Gets the specified property.
+
+    Specify an individual jail by its name or use `defaults` to get the hosts
+    defaults from the main source dataset.
+    """
 )
 @click.pass_context
 @click.argument("prop", nargs=-1, required=False, default=None)
--- a/iocage_cli/get.py
+++ b/iocage_cli/get.py
@@ -1,4 +1,5 @@
 # Copyright (c) 2014-2018, iocage
+# Copyright (c) 2017-2018, Stefan Grönke
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


iocage_cli/get.py
```diff
@@ -21,120 +22,122 @@
...
-        })
-        exit()
+    logger = ctx.parent.logger
+    host = iocage.Host.Host(logger=logger)
+
+    _prop = None if len(prop) == 0 else prop[0]
+
+    if _pool is True:
+        try:
+            print(host.datasets.active_pool.name)
```
igalic: is this all pools? activated in rc.conf or only the one in zfs?

iocage_cli/get.py
```diff
@@ -21,120 +22,122 @@
...
+
+    if _all is True:
+        if jail is None:
+            jail = _prop
+        _prop = None
+
+    if _prop == "all":
+        _prop = None
+
+    if jail == "defaults":
```
igalic, line 74: this needs documentation in the help

iocage_cli/get.py
```diff
@@ -21,120 +22,122 @@
...
+            exit(1)
+
+    for key in source_resource.config.all_properties:
+        if (_prop is None) or (key == _prop):
+            value = source_resource.config.get_string(key)
+            _print_property(key, value)
+
+
+def _print_property(key: str, value: str) -> None:
+    print(f"{key}:{value}")
```
igalic, line 113: can we get this in --output=json format?

iocage_cli/set.py
```diff
@@ -21,39 +22,83 @@
...
+            target=host.defaults
+        )
+        if len(updated_properties) > 0:
+            logger.screen("Defaults updated: " + ", ".join(updated_properties))
+        else:
+            logger.screen("Defaults unchanged")
+        return
+
+    # Jail Properties
+    filters = (f"name={jail}",)
```
igalic, line 71: is this currently able to match iocage/iocage's jail's ***and*** templates?

igalic (commented):
👀

skarekrow:
This is lacking most flags iocage/iocage's `get` currently supports, and as @igalic mentioned getting libxo output like we have with v2's list would be great

iocage_cli/set.py
gronke: No, jails stored in the templates dataset are not seen by libiocage at the moment. It will be the case once https://github.com/iocage/libiocage/issues/532 is resolved without requiring changes to the CLI.

iocage_cli/get.py
gronke: You can get that via `iocage list --output-format=json` <myjail>`. I'd prefer to add the enhancement after finishing the v2 transition.

@igalic please feel free to create an issue for that that points to Milestone 2.0.

gronke:
> This is lacking most flags iocage/iocage's `get` currently supports, and as @igalic mentioned getting libxo output like we have with v2's list would be great

- It is already possible to query jail state via the get command `iocage get jid <myjail>`
- Plugins are further decoupled from iocage and are available as Provisioning Module called `ix`. This will be one of the later merges, so we need to be patient before the plugin support re-appears at a different place.
- Indeed, information about pools and their state is important. Will create an Issue that reminds us to implement listing pools (yes, multiple of them).

iocage_cli/get.py
igalic: https://github.com/iocage/iocage/issues/639

skarekrow:
Let's keep this open until #3 is done in your list then, we can always add #2 later.

gronke:
> Let's keep this open until #3 is done in your list then, we can always add #2 later.

@skarekrow can you please put in the correct PR or Issue references?

skarekrow:
1. It is already possible to query jail state via the get command iocage get jid <myjail>
2. Plugins are further decoupled from iocage and are available as Provisioning Module called ix. This will be one of the later merges, so we need to be patient before the plugin support re-appears at a different place.
3. Indeed, information about pools and their state is important. Will create an Issue that reminds us to implement listing pools (yes, multiple of them).

gronke:
> Let's keep this open until #3 is done in your list then, we can always add #2 later.

@skarekrow It's implemented 👋 This PR adds `iocage list --pools` that list all active pools.

iocage_cli/get.py
gronke: Good catch. This feature needs to be removed because it has been replaced by `iocage list -p`.

iocage_cli/get.py
```diff
@@ -37,7 +37,11 @@
 @click.command(
     context_settings=dict(max_content_width=400,),
     name="get",
-    help="Gets the specified property."
+    help="""Gets the specified property.
+
+    Specify an individual jail by its name or use `defaults` to get the hosts
```
igalic: > the `host's`

igalic (commented):
grammar

iocage_cli/get.py
gronke: Left a hint in the Click commands global heading. According to the Click documentation arguments are supposed to be described there.

skarekrow (approved):

PR #637, opened 2018-10-05
libiocage: import and export commands

gronke:
- copy import and export commands from libiocage/ioc

### DIFF
--- a/iocage_cli/export.py
+++ b/iocage_cli/export.py
@@ -1,4 +1,5 @@
 # Copyright (c) 2014-2018, iocage
+# Copyright (c) 2017-2018, Stefan Grönke
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


iocage_cli/export.py
```diff
@@ -21,16 +22,75 @@
...
+    recursive = False
+
+    ioc_jail = iocage.Jail.JailGenerator(
+        jail,
+        logger=logger,
+        zfs=zfs,
+        host=host
+    )
+
+    if os.path.isfile(destination) is True:
```
william-gr: Should this be os.path.exists ? In case someone try using a directory or something that is not a regular file.

gronke: Yes, that's indeed undefined here. When it is a directory, would you expect some automatically generated file name or to see the same error message?

william-gr: In my opinion the same error message would be OK

skarekrow (approved):

iocage_cli/export.py
```diff
@@ -21,16 +22,75 @@
...
+    "-s", "--standalone",
+    default=False,
+    is_flag=True,
+    help="Exports the jails root dataset independently"
+)
+# @click.option(
+#     "-r", "--recursive",
+#     default=False,
+#     is_flag=True,
+#     help="Include ZFS snapshots. Implies --standalone"
```
igalic, line 56: why is this commented out?

iocage_cli/export.py
```diff
@@ -21,16 +22,75 @@
...
+    ctx: IocageClickContext,
+    jail: str,
+    destination: str,
+    standalone: bool
+) -> None:
+    """
+    Backup a jail.
+
+    The selected jail will be exported to a gzip compressed tar archive stored
+    as the destination path.
```
igalic, line 68: @_@

half of docker :P

iocage_cli/import.py
```diff
@@ -21,16 +22,47 @@
...
+@click.argument("source", required=True)
+def cli(
+    ctx: IocageClickContext,
+    jail: str,
+    source: str
+) -> None:
+    """Restore a jail from a backup archive."""
+    logger = ctx.parent.logger
+    zfs: iocage.ZFS.ZFS = ctx.parent.zfs
+    host: iocage.Host.HostGenerator = ctx.parent.host
```
igalic, line 50: does this mean you can import from a *remote* host?

igalic (commented):
👀

iocage_cli/import.py
gronke: Yes, of course! That's the whole purpose of a backup export 🙊

iocage_cli/export.py
gronke: It is disabled because it currently cannot be imported but will be enabled in a later step. I don't want users to create backups they have to manually import later on.

PR #642, opened 2018-10-06
Large rework to Exec interface

skarekrow:
iocage side of #37742, still requires plugin work in freenas/freenas.

- Use python dns.resolver
- Drop other types of DNS tests
- Make Exec a generator
- New Exception type CommandFailed for Exec failures
- Use Exec for RELEASE updating along with jail
- Break console out of Exec
- Log update progress now (means colors as well)

Ticket: #37742

### DIFF
--- a/iocage_lib/ioc_exec.py
+++ b/iocage_lib/ioc_exec.py
@@ -28,6 +28,7 @@ import iocage_lib.ioc_common
 import iocage_lib.ioc_json
 import iocage_lib.ioc_list
 import iocage_lib.ioc_start
+import iocage_lib.ioc_exceptions
 import select
 import fcntl
 import os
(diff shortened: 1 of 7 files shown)


iocage_lib/ioc_exec.py
```diff
@@ -83,136 +83,173 @@ def exec_jail(self):
...
+                # Courtesy of @william-gr
+                # service(8) and some rc.d scripts have the bad habit of
+                # exec'ing and never closing stdout/stderr. This makes
+                # sure we read only enough until the command exits and do
+                # not wait on the pipe to close on the other end.
+                #
+                # Same issue can be demonstrated with:
+                # $ jexec 1 service postgresql onerestart | cat
+                # ... <hangs>
+                # postgresql rc.d command never closes the pipe
```
gronke, line 169: A good one! I had the same experience with graylog and even had to work with pseudo TTYs to make it start correctly.

gronke (approved):

gronke (changes requested):
`import dns` fails because the dependency is not installed. You should add `dnspython` to the list of Python dependencies.

iocage_lib/ioc_exec.py
igalic: shouldn't these be submitted as bugs somewhere?

gronke: Seems we have worked around the issue on all ends. But yes, I find it quite unfortunate to rely on a TTY when starting services. We should bring this up and can decide to write a script that tests for similar issues with other services.

skarekrow:
Wrong target repo

PR #644, opened 2018-10-06
libiocage: Deprecated CLI Commands

gronke:
The `rollback` command is now covered by the `snapshot` command itself.

```shell-session
$ python3 . snapshot --help                                                                                              :(
Usage: iocage snapshot [OPTIONS] COMMAND [ARGS]...

  Take and manage resource snapshots.

Options:
  --help  Show this message and exit.

Commands:
  list      List all snapshots
  create    Create a snapshot
  rollback  Rollback to a snapshot
  remove    Delete existing snapshots
```


igalic:
i was gonna say, given libiocage's stability status, we can simply delete the file, then i looked at the diff 😅

igalic (approved):
👍

skarekrow:
Why the removal of df and debug? What are the replacements?

skarekrow (changes requested):
Should have selected request changes on this :D

igalic:
for debug this is the idea: https://github.com/iocage/libiocage/issues/214

as far as `df` goes, `zfs` has that built in see https://github.com/iocage/libiocage/issues/207

skarekrow:
The first isn't even implemented, so why deprecate something that has no replacement and is highly useful in figuring out problems? Besides, FreeNAS relies on that, and that being removed is a showstopper for us.

As for df, once again, neither of those commands you specified later in that give the *same* output without more tricks (the compressratio, reservation and quota immediately come to mind) so while the user *can* do these things, they can also do everything iocage does themselves as well. The idea is to make life easier by doing it for them ;)

So tldr; I do not agree with either of these being removed.

gronke:
> The first isn't even implemented, so why deprecate something that has no replacement and is highly useful in figuring out problems? Besides, FreeNAS relies on that, and that being removed is a showstopper for us.

ZFS can generate this output without needing iocage to do it:

```sh
zfs get compressratio,reservation,quota,used,available <DATASET>
```

The dataset can be derived from `ioc get dataset_name <JAIL>`, so that both combined generates the desired output:

```
$ zfs get compressratio,reservation,quota,used,available `ioc get dataset_name myjail`
NAME                                       PROPERTY       VALUE   SOURCE
zroot/iocage/jails/myjail  compressratio  2.03x   -
zroot/iocage/jails/myjail  reservation    none    default
zroot/iocage/jails/myjail  quota          none    default
zroot/iocage/jails/myjail  used           327M    -
zroot/iocage/jails/myjail  available      241G    -
```

> As for df, once again, neither of those commands you specified later in that give the same output without more tricks (the compressratio, reservation and quota immediately come to mind) so while the user can do these things, they can also do everything iocage does themselves as well. The idea is to make life easier by doing it for them ;)

The implementation that was deprecated from the CLI here is rather static. It does not give users much flexibility to choose the output. Since all this is already implemented in ZFS, why not use the dataset output of a jail to obtail all required information. I don't find it necessary to reinvent the interfaces of zfs get to achieve the same.

skarekrow:
`df` can likely be removed, but it's just another step for a user. I disagree with that line of thinking. We're a jail _manager_. The point being to make things easy and convenient. Just because the user *can* do something in a different method, doesn't mean we can't wrap it and provide a convenient interface. If that was our line of thinking, a lot of things could be dropped from iocage, possibly including the project. Why have `stop` when the user can simple call `iocage exec <JAIL> /etc/rc.shutdown` and manually remove the JID. Why have `exec` when the user can just use `jexec`, etc.

It's not very user-friendly is the key point.

As for `debug`, the zfs string you mentioned isn't related, have you ran `debug`? ZFS alone cannot generate that output :P It is very helpful in debugging, we should never deprecate things a user can run easily to give us more insight into issues.

PR #646, opened 2018-10-08
libiocage: migrate command

gronke:
- Copy migrate command from libiocage/ioc


gronke:
Wrong target branch - should be v2.

PR #647, opened 2018-10-08
libiocage: update command

gronke:
- Deprecate upgrade command
- Copy update command from libiocage/ioc


gronke:
Wrong target branch - should be v2.

PR #649, opened 2018-10-08
libiocage: update command

gronke:
- Deprecate upgrade command
- Copy update command from libiocage/ioc


skarekrow:
Why the removal of upgrade? Update != Upgrade in FreeBSD after all.

skarekrow (changes requested):
Should have selected request changes on this :D

gronke:
> Why the removal of upgrade? Update != Upgrade in FreeBSD after all.

It is solved in the update command logic. It applies the updates that have been fetched for the release, but it is only necessary if it is not a basejail. The `fetch` command downloads the updates while `update` installs them to a target jail.

After updating a jail it is still recommended to update the packages and run mergemaster, but this ~~is~~ can be a manual process.

skarekrow:
I don't think we're talking about the same thing. Updating is patch level. Upgrading is moving from another world to a new one. 

IE 11.1-RELEASE -> 11.2-RELEASE
Update is 11.1-RELEASE -> 11.1-RELEASE-p4

If we are talking about the same thing, you must not confuse those terms, hence the separate commands. They are very different with different goals.

PR #650, opened 2018-10-08
libiocage: migrate command

gronke:
- Copy migrate command from libiocage/ioc

### DIFF
--- a/iocage_cli/migrate.py
+++ b/iocage_cli/migrate.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2014-2018, iocage
+# Copyright (c) 2014-2017, iocage
 # All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


skarekrow (approved):

iocage_cli/migrate.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2014-2018, iocage
+# Copyright (c) 2014-2017, iocage
```
igalic, line 1: eh?

igalic (commented):
©

iocage_cli/migrate.py
skarekrow: Bah haha good catch.

PR #651, opened 2018-10-09
Fixed building py-libzfs after updating

thedix:
py-libzfs added "configure" command in the latest version and iocage build fails without calling it.


gronke:
@william-gr does this have an impact on the ports package?

gronke (approved):

skarekrow:
Thanks!

thedix:
> @william-gr does this have an impact on the ports package?

New py-libzfs compiles from ports on 11.2 without any problems.

PR #653, opened 2018-10-09
iocage console hangs

jsegaert:

- [x] Explain the feature

It needs a `return` here; when exiting out of "iocage console", there no need to execute the next code block.  In fact, it makes the terminal client hang.


skarekrow (approved):

skarekrow:
Great catch!

Issue #521, opened 2018-04-13
Do not add default gw to bridge on iocage start


ole-db:
Make sure to follow and check these boxes before submitting an issue! Thank you.

- [x] Supply `iocage --version`
Version	0.9.10 2017/12/22

if I start a Jail with vnet Interface the interface with the default route on it ( in my case `em0`) is automatically added to the bridge.

In my point of view, this makes sense in some special setups. There are many setups, where you don't want to have this as a standard behaviour. Or need a way to deactivate.

https://github.com/iocage/iocage/blob/1908813cbd01c6770ced228f4c20c8ba80f3dad9/iocage/lib/ioc_start.py#L605-L615

gronke:
> if I start a Jail with vnet Interface the interface with the default route on it ( in my case em0) is automatically added to the bridge.

This is a security risk and should not be done by iocage. One can end up exposing critical infrastructure to the public by accident or getting a root server banned for ARPing on the uplink interface.

At the most this action should only be performed when users opt-in to use it, but I'm in favor of printing a warning message instead that may suggests the command to execute to the device.

skarekrow:
I'd be willing to accept a PR, but this was the previous behavior in iocage-legacy and is ported as a POLA type approach. That being said, this is a very common requirement. You cannot add the vnet interfaces to a bridge without having a host interface for them to route packets out.

I'd be willing to see a more concrete example of what you think it should do instead, but I'm not able to dedicate any time to implementing anything currently.

ole-db:
Ok. Bridging all vnet interfaces whit the host interface can be used in very simple setups, I agree. But only in Setups where all Jails are in the same Layer2 subnet with the host. But in all my personal use cases this is not true. Let me give you two examples:

1.
At many cheap hosting providers you can rent a server with only one IPv4 address. If you need more Addresses you can rent several /2x networks. These networks are routed to your server. In this case it makes no sense to add your host interface to the bridge. Only vnet interfaces are member of the bridge and you set `sysctl net.inet.ip.forwarding=1`

2.
Your hosting server and your Jails are in different vlans. in this case you want to add the vlan interface of your jails network to the bridge.

my suggestion is to delete line 605 to line 615. In my opinion iocage should not do this network setup. This should be placed in /etc/rc.conf 

I could make a PR, but at first I have to figure out how to make multiple PR in one project. sorry I am new to github ;-)

skarekrow:
Valid points! Though for point 2 you would also need your host interface
[...]

PR #655, opened 2018-10-12
if 'vnet_default_interface' is set to 'none' no host interface will be added to the bridge

ole-db:
if 'vnet_default_interface' is set to 'none' no host interface will be added to the bridge. 
Auto-detected interface is added to bridge if 'vnet_default_interface' is set to 'auto'

please see https://github.com/iocage/iocage/issues/521

### DIFF
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -1007,11 +1007,7 @@ class IOCStart(object):
                 if default_if == 'auto':
                     default_if = self.get_default_gateway()[1]
 
-                if default_if == 'none':
-                    bridge_cmd = [
-                        "ifconfig", bridge, "create"
-                    ]
-                else:
+                if default_if != 'none':
                     bridge_cmd = [
                         "ifconfig", bridge, "create", "addm", default_if
                     ]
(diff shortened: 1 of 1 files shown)


gronke (dismissed):
This changes mitigate the issue described in #521.

I still do not get why iocage should create bridges at all.

iocage_lib/ioc_start.py
```diff
@@ -1002,12 +1004,18 @@ def find_bridge_mtu(self, bridge):
...
                 # Let's get the default vnet interface
                 default_if = self.get('vnet_default_interface')
-                if default_if == 'none':
+                if default_if == 'auto':
                     default_if = self.get_default_gateway()[1]
 
-                bridge_cmd = [
-                    "ifconfig", bridge, "create", "addm", default_if
-                ]
+                if default_if == 'none':
```
skarekrow: Let's change this to `if default_if != 'none':` and drop the empty bridge creation, keeping only the else block under this.

skarekrow (changes requested):
In this context `auto` == `none`. So we don't need most of the PR here, However we should not be adding any interface to a bridge if the property is set to `none`, as I noted in the PR.

skarekrow (approved):

skarekrow:
Thanks @ole-db!

PR #672, opened 2018-10-30
Quietly start jail if needed when running 'iocage exec'

### LINKED ISSUE #668: Running 'iocage exec' when jail is down

 a stable release.

On iocage master, when running `iocage exec`, the command fails when the jail is not running.
```
➜  ~ sudo iocage exec foo cat /etc/motd
jexec: jail "ioc-foo" not found
Command failed!
Exception: "CommandFailed:Command: ('cat', '/etc/motd'), stdout: None, stderr: None" occured
```
If I remember correctly, I think it used to automatically start the jail.  Was that done on purpose, or is that a regression?

jsegaert:

- [X] Explain the feature

This fixes #668

### DIFF
--- a/iocage_lib/iocage.py
+++ b/iocage_lib/iocage.py
@@ -833,8 +833,23 @@ class IOCage(object):
                 ) + command
 
             try:
+                _started = False
+                _silent = self.silent
+                _status, _ = self.list("jid", uuid=uuid)
+
+                if not _status:
+                    self.silent = True
+                    self.start()
+                    _started = True
+                    self.silent = _silent
+
                 su_command = su.Popen(interactive_cmd, env=su_env)
                 stdout, stderr = su_command.communicate()
+
+                if _started:
+                    self.silent = True
+                    self.stop()
+                    self.silent = _silent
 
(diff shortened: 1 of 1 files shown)


skarekrow (commented):
Are we sure this is only needed for interactive commands?

gronke (changes requested):
This behavior matches with libiocage `exec --fork`. 👍😍 

I'd love to see the `-f/--fork` option added, so that the CLI commands use a common syntax.

jsegaert:
Great question.  Correct; only the interactive commands are not covered.  Similar logic already exists for "pkg" and "console" and the IOCExec generator does it in the "flight_checks".

Note that because of the latter, some commands such as `iocage get -P` require root privileges because they may attempt to start the jail (and would need it to run jexec anyway).

skarekrow:
@gronke That doesn't make sense, these commands aren't forking. They're starting a jail that isn't running so the command can be run. That's a separate PR if we want to add that to v1. The user could do it manually by adding `&` to their command and the jail would return immediately anyways.

skarekrow (approved):

PR #673, opened 2018-10-30
fixes confusion with on/off and yes/no

ole-db:
too many boolean types ...

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -1165,7 +1165,7 @@ class IOCJson(object):
             "owner": ("string", ),
             "priority": str(tuple(range(1, 100))),
             "hostid": ("string", ),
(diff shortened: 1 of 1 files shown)


skarekrow:
I don’t agree with this change. Considering dhcp still has a different boolean, and this only changed the property. We have to change the usage in the program as well. 

 If anything we should change all booleans to accept off/on and yes/no.

ole-db:
I don't understand.

This is a bug fix. 'hostid_strict_check' has to be 'off' or 'on'. Otherwise it will not take effect. Without this fix you can't use the 'iocage set' commad to set 'hostid_strict_check' to 'off' or 'on'.

skarekrow:
I guess that proves the confusion xD thanks for the PR

skarekrow (approved):

PR #675, opened 2018-11-01
Fixed sorting jails at iocage get -r

thedix:
Fixed sorting jails by name when calling recursive "iocage get".


skarekrow (approved):

skarekrow:
Thanks!

PR #681, opened 2018-11-09
Optionally allow mounting fusefs from within the jail.

asomers:
Add a new parameter that allows mounting fuse filesystems within the jail.  Requires FreeBSD 12.0 or newer on the host.  Tested with fusefs-encfs, a 13.0-CURRENT host, and an 11.2-RELEASE guest.

- [x] Explain the feature

### DIFF
--- a/iocage.8
+++ b/iocage.8
@@ -1540,6 +1540,18 @@ Default: 0
 .Pp
 Source:
 .Xr jail 8
+.It Pf allow_mount_fusefs= Op 0 | 1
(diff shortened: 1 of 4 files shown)


jsegaert:
This will break iocage running on hosts < 12.0.  Please consider adding some logic to prevent this, as e.g. the `allow_mlock` flag does.

asomers:
Here you go @jsegaert

skarekrow (approved):

skarekrow:
Thanks guys!

PR #685, opened 2018-11-12
Fixed error message for corrupted default.json (ticket 56364).

a-stjohn:
Tested change in VM: No issues.

Do I also need to add this change in freenas/iocage/... repo?


skarekrow (approved):

skarekrow:
Yep make a PR in freenas/iocage cherry picking this commit from master.

PR #686, opened 2018-11-13
Drop "zero dependency" from description

utrenkner:
This still stemmed from the old shell based iocage.

No further comment/explanation needed here, I guess...

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -7,7 +7,7 @@
 iocage - A FreeBSD Jail Manager
 ===============================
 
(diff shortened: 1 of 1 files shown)


skarekrow (changes requested):
Ha! Good catch. Though we should remove drop in as well. Python made that part very difficult.

utrenkner:
Yes, makes sense! I removed "drop in" as well.

skarekrow (approved):

skarekrow:
Thanks!

Issue #700, opened 2018-11-29
Problem with ZFS properties and partial config


thedix:
I use the latest version from master.
The problem: unable to set zfs properties for newly created jails:
```
# iocage set quota=20G test1
quota is not a valid property!
```
The problem is caused by [this line](https://github.com/iocage/iocage/blob/master/iocage_lib/ioc_json.py#L1291).
At that moment variable conf has the following content:
`{'allow_raw_sockets': '1', 'basejail': 'yes', 'cloned_release': '11.2-RELEASE', 'host_hostname': 'test1', 'host_hostuuid': 'test1', 'jail_zfs_dataset': 'iocage/jails/test1/data', 'release': '11.2-RELEASE-p4', 'quota': '20G'}`
So conf["template"] == "yes" check fails because there is no such key "template".
And exception is raised "not a valid property".

I think this bug is caused by new partial config for jails.
I'd fix this bug if it were a small typo.
But seems changing full config to a partial config could be a serious issue because there are plenty of places with direct checking config via "conf['varname']".
This bug affects new jails only, old jails created with full config are not affected.

@skarekrow, could you explain reasons to move to partial config?
Yes, it looks nice, but there are some implications:
1. With partial config it is not comfortable to make config changes directly to config.json. For example, I'd like to enable raw sockets for jail and I don't remember the exact variable name. I open config.json, search for "raw", find "allow_raw_sockets", and set it to 1.
2. Default config can be changed in the future versions (yes!). And when migrating from host with older iocage version to a host with newer version (different defaults.json) could lead to all jails change their behavior unexpectedly.
3. As I mention before, there are a lot of places in the code with direct array access to conf. This can be fixed by using some function to get values, but this should be done in the whole code - is it real?

skarekrow:
In this case the bug is a small typo, using just `.get` would fix it. There are unfortunately some spots that may have been missed, and as you noted those have implications. Let's address your concerns here:

1 - Just open the defaults.json to search for that name, or `man iocage`. The idea is to make it easier to tell exactly *what* was changed from the defaults per jail.
2 - That's true, so if you rely on their behavior I'd suggest migrating defaults.json, or setting those properties per jail. I suppose we could use a `create` flag to use "thick" configurations to bypass this behavior, allowing the user to use the defaults(thin configuration) unless they have good reason not too (your use-case)
[...]

PR #701, opened 2018-11-29
Fixed bug #700: get non-existing keys from conf

thedix:
This fixes bug #700 and other potential bugs with vnet.


thedix:
Funny numbers: bug #700 and pull request #701 :)

thedix:
Is it ok to fix this way?

skarekrow (approved):
For now yes I think this is a good fix. Will think on what you said in the issue and figure out a better global solution.

PR #708, opened 2018-12-04
Create multiple jails when uuid is supplied as name

sonicaj:
This commit fixes a bug where we were not able to create count jails when a valid uuid was supplied as a name. Apart from that this commit enforces that count flag cannot be used with uuid flag and removes force option from create command as it is not used anywhere.
Ticket: #61551

### DIFF
--- a/iocage_cli/create.py
+++ b/iocage_cli/create.py
@@ -25,6 +25,7 @@
 import json
 import os
 import re
+import uuid
 
 import click
 import iocage_lib.ioc_common as ioc_common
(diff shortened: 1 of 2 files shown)


iocage_cli/create.py
```diff
@@ -81,11 +82,25 @@ def validate_count(ctx, param, value):
...
                    " 36.")
-@click.option("--force", "-f", is_flag=True, default=False,
-              help="Skip the interactive question.")
 @click.argument("props", nargs=-1)
 def cli(release, template, count, props, pkglist, basejail, thickjail, empty,
-        short, name, _uuid, force, thickconfig):
+        short, name, _uuid, thickconfig):
+
+    if _uuid:
+
```
skarekrow: Remove extra line

skarekrow (changes requested):

skarekrow (approved):

PR #714, opened 2018-12-10
Bug fix for template jails

sonicaj:
This commit fixes a bug which prevented creating a template jail at the same time the jail was being created for the first time. The dataset was readonly at that point and system tried to complete the relevant configuration files but failed to do so because of the readonly property set to true.
Ticket: #62190

### DIFF
--- a/iocage_lib/ioc_create.py
+++ b/iocage_lib/ioc_create.py
@@ -416,6 +416,8 @@ class IOCCreate(object):
                 exit(1)
 
         if not self.plugin:
+            # TODO: Should we probably only write once and maybe at the end
+            # of the function ?
             iocjson.json_write(config)
 
         # Just "touch" the fstab file, since it won't exist and write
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_create.py
```diff
@@ -436,6 +438,12 @@ def _create_jail(self, jail_uuid, location):
             jail_uuid_short = jail_uuid
             jail_hostname = jail_uuid
 
+        # If jail is template, the dataset would be readonly at this point
+        if is_template:
```
skarekrow, line 442: You need to set it back to readonly after we're done writing

iocage_lib/ioc_create.py
```diff
@@ -416,6 +416,8 @@ def _create_jail(self, jail_uuid, location):
                 exit(1)
 
         if not self.plugin:
+            # TODO: Should we probably only write once and maybe at the end
```
skarekrow, line 419: Ideally! It would involve a fair bit of churn though.

skarekrow (changes requested):

iocage_lib/ioc_create.py
sonicaj: We already do that at the very end of this function ;)

iocage_lib/ioc_create.py
sonicaj: If you think I should go ahead with that, I can have the changes in this PR and you can review it then ?

skarekrow (approved):

PR #718, opened 2018-12-13
Check snapshot exists before destroying

sonicaj:
This commit ensures that we check a snapshot exists before we destroy it and log an appropriate message accordingly.
Ticket: #63801


skarekrow (changes requested):
Whoops forgot this one! We just changed IOCZFS to use subprocess, so let's rewrite this to use that method instead.

sonicaj:
I have moved this in another PR ;)

PR #729, opened 2018-12-17
tkt-64638: Fix mismatching RELEASEs being fetched, and a couple other bugs

skarekrow:
Noted by @jsegaert

- Use the right interface for fetching instead of internal plumbing
- Fix bad variable usage
- Fix a couple other exceptions that could occur with the fetch interface

FreeNAS Ticket: #64638


skarekrow:
@jsegaert ^

william-gr (approved):

PR #731, opened 2018-12-18
tkt-64899: Rewrite IOCZFS to use subprocess calls

skarekrow:
libzfs writes to stderr, and messing with fd's is not very clean. This also seems to perform better

FreeNAS Ticket: #64899

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -67,7 +67,7 @@ class IOCZFS(object):
             try:
                 p, v = prop.split()
             except ValueError:
-                v = '-'
+                p, v = prop.strip(), '-'
 
             p_dict[p] = v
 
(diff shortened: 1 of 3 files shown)


iocage_lib/ioc_json.py
```diff
@@ -50,36 +50,56 @@ def __init__(self):
...
+                'all',
+                identifier
+            ], stdout=su.PIPE, stderr=su.PIPE
+        ).stdout.decode().splitlines()
+
+        for prop in props:
+            try:
+                p, v = prop.split()
+            except ValueError:
+                v = '-'
```
william-gr: Where is `p`?

iocage_lib/ioc_json.py
```diff
@@ -50,36 +50,56 @@ def __init__(self):
...
+            su.run(
+                [
+                    'zfs', 'set', f'{key}={value}', identifier
+                ], stdout=su.PIPE
+            )
         else:
-            ds[key].value = value
+            su.run(
+                [
+                    'zpool', 'set', f'{key}={value}', identifier
```
william-gr: Why would a set property use zpool set?

william-gr (changes requested):

william-gr (approved):

PR #734, opened 2018-12-18
tkt-62547: No longer start network services explicitly

### LINKED ISSUE #711: iocage should not start services in the jail on behalf of the admin

Version	1.0 ALPHA 1 (FreeNAS 11.2-RELEASE)
Create jail with IPv6 autoconfiguration and VNET. Intentionally remove `rtsold_enable` entry from `/etc/rc.conf`in the jail. Restart jail.
```
root@freenas-pmh:~ # iocage console cloud
root@cloud:~ # service rtsold status
Cannot 'status' rtsold. Set rtsold_enable to YES in /etc/rc.conf or use 'onestatus' instead of 'status'.
root@cloud:~ # ps awux|grep rtsold
root       94984  0.0  0.0    6312   2056  -  IsJ  14:36   0:00.00 /usr/sbin/rtsold -a
root        6518  0.0  0.0    6660   2464  2  S+J  16:17   0:00.00 grep rtsold
```
See above.
iocage should not start rtsold if the admin explicitly disabled in inside the jail - big POLA violation. It is of course great to provide reasonable defaults, but after jail creation iocage should not mess with things *inside* *ever*
- [ ] Checked that the problem has not already been fixed on `master` if using
 a stable release.

Kind regards
Patrick

skarekrow:
- Use explicit fstab for sysrc during creation
- New syntax for ip4_addr
EXAMPLE : ip4_addr='vnet0|DHCP' 
This replaces dhcp=on, as that previous property would start DHCP on all interfaces, which isn't always desired.
- Simplify setting rc.conf in start to use sysrc for DHCP or rtsold.
- New property 'rtsold', this will enable the service in the jail. This requires having accept_rtadv in the ip6_addr, this new property allows one to rely on just rtsol instead of always starting rtsold.

Closes #711
FreeNAS Ticket: #62547

### DIFF
--- a/iocage_lib/ioc_create.py
+++ b/iocage_lib/ioc_create.py
@@ -374,7 +374,9 @@ class IOCCreate(object):
                         config['vnet'] = 'on'
 
                     rtsold_enable = 'YES'
-            elif key == 'dhcp' and value == 'on':
+            elif (key == 'dhcp' and value == 'on') or (
+                key == 'ip4_addr' and 'DHCP' in value
+            ):
                 if 'vnet=on' not in self.props:
                     iocage_lib.ioc_common.logit({
                         'level': 'WARNING',
@@ -889,8 +891,8 @@ ipv6_activate_all_interfaces=\"YES\"
             su.Popen(
                 ['mount', '-F', f'{location}/fstab', '-a']).communicate()
 
-        su.Popen(["sysrc", "-R", f"{location}/root",
-                  f"hostname={host_hostname}"],
+        su.Popen(['sysrc', '-f', f'{location}/root/etc/rc.conf',
+                  f'hostname={host_hostname}'],
                  stdout=su.PIPE).communicate()
 
         if basejail != 'no':
--- a/iocage_lib/ioc_create.py
+++ b/iocage_lib/ioc_create.py
@@ -375,7 +375,7 @@ class IOCCreate(object):
 
                     rtsold_enable = 'YES'
             elif (key == 'dhcp' and value == 'on') or (
-                key == 'ip4_addr' and 'DHCP' in value
+                key == 'ip4_addr' and 'DHCP' in value.upper()
             ):
                 if 'vnet=on' not in self.props:
                     iocage_lib.ioc_common.logit({
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -143,7 +143,7 @@ class IOCStart(object):
         dhcp = self.conf["dhcp"]
         rtsold = self.conf['rtsold']
         wants_dhcp = True if dhcp == 'on' or 'DHCP' in self.conf[
-            'ip4_addr'] else False
+            'ip4_addr'].upper() else False
         vnet_interfaces = self.conf["vnet_interfaces"]
         ip6_addr = self.conf["ip6_addr"]
         prop_missing = False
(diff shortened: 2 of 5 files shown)


skarekrow:
@pmhausen Hey can you test this PR to see if this satisfies your rtsold requirements? Thanks!

iocage_lib/ioc_create.py
```diff
@@ -374,7 +374,9 @@ def _create_jail(self, jail_uuid, location):
                         config['vnet'] = 'on'
 
                     rtsold_enable = 'YES'
-            elif key == 'dhcp' and value == 'on':
+            elif (key == 'dhcp' and value == 'on') or (
+                key == 'ip4_addr' and 'DHCP' in value
```
william-gr: Requiring DHCP uppercase seems odd. Is there anything which forces DHCP to be set as uppercase?

Perhaps

'dhcp' in value.lower() would make sense?

iocage_lib/ioc_start.py
```diff
@@ -982,9 +1000,11 @@ def find_bridge_mtu(self, bridge):
             dhcp = 'off'
```
sonicaj, line 1001: This will fail unit tests as `wants_dhcp` would not exist

iocage_lib/ioc_start.py
```diff
@@ -930,29 +928,49 @@ def __start_generate_vnet_mac__(self, nic):
...
-            if not replaced:
-                # They didn't have any interface in their rc.conf,
-                # fresh jail perhaps?
-                _rc.insert(0, f'ifconfig_{nic}="DHCP"\n')
+    def __check_rtsold__(self):
+        if 'accept_rtadv' not in self.conf['ip6_addr']:
+            iocage_lib.ioc_common.logit(
+                {
+                    'level': 'EXCEPTION',
+                    'message': 'Must set atleast one ip6_addr to accept_rtadv!'
```
william-gr: atleast

iocage_lib/ioc_start.py
```diff
@@ -586,6 +540,59 @@ def __start_jail__(self):
                 _callback=self.callback,
                 silent=self.silent)
 
+            if not vnet_err and vnet and wants_dhcp:
+                failed_dhcp = False
+
+                try:
+                    interface = self.conf["interfaces"].split(",")[0].split(
+                        ":")[0]
```
william-gr: You are not making sense to my OCD. You're switching a bunch of double quotes to single quotes in a lot of PRs. And in this new entire code block you're using double quotes? :P

sonicaj (changes requested):
There are a few more usages as well where we might break a few other things, referencing them after this comment

sonicaj:
How should we handle listing in this case ?
https://github.com/iocage/iocage/blob/548a78681e2c7ac149735e243bb23519bed961e2/iocage_lib/ioc_list.py#L129

sonicaj:
Not sure if this is breaking or not
https://github.com/iocage/iocage/blob/548a78681e2c7ac149735e243bb23519bed961e2/iocage_lib/ioc_plugin.py#L526

sonicaj:
Would this require changing as well ?
https://github.com/iocage/iocage/blob/548a78681e2c7ac149735e243bb23519bed961e2/iocage_lib/ioc_plugin.py#L595

iocage_lib/ioc_create.py
skarekrow: DHCP is to mimic the rc.conf entry, since that makes most sense. Also nice and easy to search for later in the code.

william-gr (approved):

sonicaj (approved):

pmhausen:
> @pmhausen Hey can you test this PR to see if this satisfies your rtsold requirements? Thanks!

Of course - as soon as it hits FreeNAS 11.2 - is that scheduled for U1, already?

Kind regards
Patrick

skarekrow:
This will be in 11.3, too invasive for U1

PR #735, opened 2018-12-19
tkt-63801: Check snapshot exists before destroying

sonicaj:
This commit ensures that we check a snapshot exists before we destroy it and log an appropriate message accordingly. It also introduces a snapshot class which we can use to better manage snapshot related operations.
Ticket: #63801

### DIFF
--- a/iocage_lib/ioc_exceptions.py
+++ b/iocage_lib/ioc_exceptions.py
@@ -68,10 +68,7 @@ class ValidationFailed(ExceptionWithMsg):
 
 
 @contextmanager
-def ignore_exceptions(
-        *exceptions, action=None, action_args=None,
-        action_kwargs=None, suppress_exception=True
-):
+def ignore_exceptions(*exceptions, return_value=None, suppress_exception=True):
     try:
         yield
     except exceptions as e:
@@ -80,10 +77,7 @@ def ignore_exceptions(
             # exceptions
             raise e
 
-        if callable(action):
-            action(
-                *(action_args or []),
-                **(action_kwargs or {})
-            )
+        if callable(return_value):
+            return return_value()
         else:
-            return action
+            return return_value
--- a/iocage_lib/ioc_exceptions.py
+++ b/iocage_lib/ioc_exceptions.py
@@ -23,6 +23,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 """Exception classes for iocage"""
 import collections
+from contextlib import contextmanager
 
 
 class ExceptionWithMsg(Exception):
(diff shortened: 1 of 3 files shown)


iocage_lib/ioc_json.py
```diff
@@ -45,7 +45,76 @@
...
+    def normalize_data(self):
+        for line in self.raw_data.splitlines()[1:]:
+            self.snapshots.append(IOCSnapshot(line))
+
+    def __iter__(self):
+        for snap in self.snapshots:
+            # FIXME: I will still hold old snapshot data, let's change this pls
+            yield snap
+
+    def get_snap(self, snap_id):
```
william-gr, line 134: Listing all snapshots to get a single one is a bad idea, specially on system with thousands of them.

iocage_lib/ioc_exceptions.py
```diff
@@ -64,3 +65,25 @@ class JailMissingConfiguration(JailMisconfigured):
 
 class ValidationFailed(ExceptionWithMsg):
     pass
+
+
+@contextmanager
+def ignore_exceptions(
+        *exceptions, action=None, action_args=None,
+        action_kwargs=None, suppress_exception=True
```
william-gr: 8 spaces indentation?

iocage_lib/ioc_json.py
```diff
@@ -45,7 +45,76 @@
...
+        if self.raw_data:
+            # Expected format
+            # ['NAME', 'USED', 'AVAIL', 'REFER', 'MOUNTPOINT']
+            self.__dict__.update({
+                k: v for k, v in zip(self.attr_list, self.raw_data.split())
+            })
+
+    def delete(self, recursive=True):
+        with ioc_exceptions.ignore_exceptions(
+                su.CalledProcessError, action=False
```
william-gr: 8 spaces indentation?

iocage_lib/ioc_json.py
```diff
@@ -45,7 +45,76 @@
...
+
+class IOCSnapshots:
+    def __init__(self):
+        self.snapshots = []
+        self.normalize_data()
+
+    @property
+    def raw_data(self):
+        with ioc_exceptions.ignore_exceptions(
+                su.CalledProcessError, action=''
```
william-gr: 8 spaces indentation?

iocage_lib/ioc_exceptions.py
```diff
@@ -64,3 +65,25 @@ class JailMissingConfiguration(JailMisconfigured):
 
 class ValidationFailed(ExceptionWithMsg):
     pass
+
+
+@contextmanager
+def ignore_exceptions(
+        *exceptions, action=None, action_args=None,
```
william-gr: action is a confusing name IMHO. The only use cases you have so far are static return values, and thats what this really is, a `return_value`

If a callable is a edge cage perhaps it makes sense to only use the callable and no args and kwargs. These could be passed using functools.partial if needed.

skarekrow (approved):

PR #736, opened 2018-12-19
tkt-59247: Add Functional Tests

sonicaj:
This PR aims to add more tests to iocage.

FreeNAS Ticket: #59247

### DIFF
--- a/tests/data_classes.py
+++ b/tests/data_classes.py
@@ -1,0 +1,643 @@
+import itertools
+import os
+import json
+import subprocess
+import uuid
+
+from datetime import datetime
+
+import libzfs
+
+
+# A helper class to parse the output of iocage for us to test
+
+class Row:
+    powers = {
+        'B': 0,
+        'K': 3,
+        'M': 6,
+        'G': 9,
+        'T': 12,
+        'P': 15
+    }
+
+    def __init__(self, raw_data, r_type=None):
+        self.raw_data = raw_data
+        for attr in [
+            'name', 'jid', 'state', 'release', 'ip4', 'ip6', 'orig_release',
+            'boot', 'type', 'template', 'basejail', 'crt', 'res', 'qta',
+            'use', 'ava', 'created', 'rsize', 'used'
+        ]:
+            setattr(self, attr, None)
+
+        assert self.raw_data is not None
+
+        if isinstance(raw_data, dict):
+            for key in raw_data:
+                setattr(self, key, raw_data[key])
+        else:
+            assert r_type is not None
+            if not hasattr(self, f'{r_type}_parse'):
+                raise NotImplemented
+
+            getattr(self, f'{r_type}_parse')()
+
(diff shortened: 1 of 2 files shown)


tests/data_classes.py
```diff
@@ -0,0 +1,647 @@
...
+
+    def df_parse(self):
+        self.name, self.crt, self.res, self.qta, self.use, self.ava = \
+            self.standard_parse()
+
+    def snapshot_parse(self):
+        self.name, self.created, self.rsize, self.used = self.standard_parse()
+
+
+class ZFS:
```
skarekrow, line 188: Perhaps we shift to subprocess zfs again starting here as well? That way we can remove our dep on it sooner and easier.

tests/data_classes.py
```diff
@@ -0,0 +1,647 @@
...
+                template = '-'
+
+        return Row({
+            'name': self.short_name if short_name else self.name,
+            'jid': self.jid,
+            'state': 'up' if self.running else 'down',
+            'boot': self.config.get('boot', 'off'),
+            'type': 'jail' if not self.is_template else 'template',
+            # TODO: Add support for plugins
+            'ip6': '-',  # FIXME: Change when ip6 tests are added
```
skarekrow, line 378: Use `'none'` to be like the other props that aren't specified
```suggestion
             'ip6': 'none',  # FIXME: Change when ip6 tests are added
```

tests/data_classes.py
```diff
@@ -0,0 +1,647 @@
...
+    def is_rcjail(self):
+        return self.config.get('boot', 'off') == 'on'
+
+    @property
+    def ip(self):
+        assert self.running is True
+        try:
+            # TODO: we should probably load interface from config and
+            # make changes if necessary to it's value then
+            # TODO: Discuss with Brandon about support for older releases
```
skarekrow, line 520: Once again, what support do you want to discuss :P

tests/data_classes.py
```diff
@@ -0,0 +1,647 @@
...
+
+            return pool['root_dataset']['properties']
+
+    def zfs_get_property(self, identifier, key):
+        try:
+            return self._zfs_get_properties(identifier)[key]['value']
+        except Exception:
+            return '-'
+
+    def get_dataset_safely(self, dataset_name):
```
skarekrow: Is there ever a time we don't want to get them safely? Perhaps just remove the other method and rename this one.

skarekrow (changes requested):

tests/data_classes.py
sonicaj: Umm this is to ensure matching output with the listing of jails etc. We don't use this for any other reason. If we change it to none, then the comparison will fail

skarekrow: Got it :+1:

skarekrow (dismissed):

skarekrow (approved):

PR #737, opened 2018-12-19
Activate and Fetch tests

sonicaj:
This commit adds tests for activate and fetch

### DIFF
--- a/tests/functional_tests/0001_activate_test.py
+++ b/tests/functional_tests/0001_activate_test.py
@@ -21,10 +21,9 @@
 # STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
 # IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
-import pytest
-from click.testing import CliRunner
 
-import iocage_cli as ioc
+import pytest
+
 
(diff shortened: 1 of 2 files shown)


tests/functional_tests/0001_activate_test.py
```diff
@@ -21,19 +21,21 @@
...
-    result = runner.invoke(ioc.cli, ['activate', zpool])
+def test_activate(zpool, invoke_cli, zfs):
+    invoke_cli(
+        ['activate', zpool]
+    )
+
+    zfs.set_pool()
 
-    assert result.exit_code == 0
+    assert zfs.pool == zpool, f'Failed to activate {zpool}'
```
skarekrow, line 41: Maybe I'm confused here, but don't you want `!=`?

tests/functional_tests/0002_fetch_test.py
```diff
@@ -21,37 +21,30 @@
...
+    command += ['-f', _file] if _file else []
+    command += ['-u', user] if user else []
+    command += ['-p', password] if password else []
+    command += ['-a', auth] if auth else []
+    command += ['-d', root_dir] if root_dir else []
+    command += ['-NU'] if noupdate else []
+
+    invoke_cli(
+        command
+    )
```
skarekrow, line 50: We should still assert this passes, it dictates all other tests after all

skarekrow (changes requested):

skarekrow (approved):

PR #739, opened 2018-12-19
tkt-65007: Return a more verbose error message on plugin failure

skarekrow:
Also, return the last 10 lines instead of just the last one.

FreeNAS Ticket: #65007


william-gr (approved):

skarekrow:
backport

PR #749, opened 2018-12-21
tkt-65610: Replace hostnames with "_" to valid hostnames

skarekrow:
Plugins that have multiple installs will run into this problem.

FreeNAS Ticket: #65610


sonicaj (approved):

skarekrow:
Unrelated test failure

PR #750, opened 2018-12-21
tkt-65628: Bug fix for iocage fetch

sonicaj:
This commit fixes a bug where we didn't fetch the host release when enter was pressed.
Ticket: #65628

### DIFF
--- a/iocage_lib/ioc_fetch.py
+++ b/iocage_lib/ioc_fetch.py
@@ -139,7 +139,7 @@ class IOCFetch(iocage_lib.ioc_json.IOCZFS):
         host_release = iocage_lib.ioc_common.get_host_release()
         for r in releases:
             message = f'[{releases.index(r)}] {r}'
-            if r in eol or []:
+            if eol is not None and (r in eol or []):
                 message += ' (EOL)'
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_fetch.py
```diff
@@ -129,13 +129,32 @@ def __fetch_eol_check__():
...
         """
         Checks if the user supplied an index number and returns the
         RELEASE. If they gave us a full RELEASE, we make sure that exists in
         the list at all.
         """
 
+        host_release = iocage_lib.ioc_common.get_host_release()
+        for r in releases:
+            message = f'[{releases.index(r)}] {r}'
+            if r in eol or []:
```
skarekrow: If `eol` is None this will cause a TypeError
```suggestion
            if eol is not None and (r in eol or []):
```

skarekrow (changes requested):

iocage_lib/ioc_fetch.py
sonicaj: Good catch 😉

skarekrow (approved):

PR #751, opened 2018-12-21
Update Makefile

sonicaj:
This commit adds a better error handling for a case where python path supplied is non-existent.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ SERVER=""
 PYTHON?=/usr/local/bin/python3
 
 install:
-	@test -s ${PYTHON} || (echo "Python path ${PYTHON} not found, please enter a valid one"; exit 1)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -3,6 +3,7 @@ SERVER=""
 PYTHON?=/usr/local/bin/python3
 
 install:
+	@test -s ${PYTHON} || (echo "Python path ${PYTHON} not found, please enter a valid one"; exit 1)
```
skarekrow:
```suggestion
 	@test -s ${PYTHON} || (echo "Python binary: ${PYTHON} not found, please enter a valid one"; exit 1)
```

skarekrow (changes requested):

skarekrow (approved):

PR #759, opened 2018-12-26
tkt-65553: Add More Tests

sonicaj:
This PR aims to introduce following changes:
1) Fixes for tests which failed earlier for jails which had clones or templates because of how snapshots happen
2) Networking Related Tests

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -88,6 +88,10 @@ def pytest_addoption(parser):
         '--upgrade', action='store_true', default=True,
         help='Decide whether or not to run upgrade tests'
     )
+    parser.addoption(
+        '--ping_ip', action='store', default='8.8.8.8',
+        help='Use --ping_ip for testing connectivity within a jail'
+    )
 
 
 def pytest_runtest_setup(item):
@@ -155,6 +159,12 @@ def dhcp(request):
 def upgrade(request):
     """Specify if upgrade test is to be executed."""
     return request.config.getoption('--upgrade')
+
+
+@pytest.fixture
+def ping_ip(request):
+    """Specify ip to be used to test connectivity within a jail"""
+    return request.config.getoption('--ping_ip')
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,6 +22,7 @@
 # IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+import ipaddress
 import os
 import re
 import subprocess
@@ -85,7 +86,7 @@ def pytest_addoption(parser):
         help='Use DHCP for creating jails'
     )
     parser.addoption(
-        '--upgrade', action='store_true', default=True,
+        '--upgrade', action='store_true', default=False,
         help='Decide whether or not to run upgrade tests'
     )
     parser.addoption(
@@ -105,7 +106,7 @@ def pytest_runtest_setup(item):
         pytest.skip('Need --dhcp option to run')
 
     if 'require_upgrade' in item.keywords and not item.config.getvalue(
-            'upgrade'
+        'upgrade'
     ):
         pytest.skip('Need --upgrade option to run')
 
@@ -124,16 +125,6 @@ def pytest_runtest_setup(item):
             )
         )
     ):
-        pytest.skip('Need either --dhcp or --jail_ip option to run')
-
-    if (
-        'require_networking' in item.keywords and all(
-            v for v in (
-                item.config.getvalue('--dhcp'),
-                item.config.getvalue('--jail_ip')
-            )
-        )
-    ):
         pytest.skip('Need either --dhcp or --jail_ip option to run, not both')
 
 
@@ -146,7 +137,19 @@ def zpool(request):
 @pytest.fixture
 def jail_ip(request):
     """Specify a jail ip to use."""
-    return request.config.getoption('--jail_ip')
+    # For tests we only support ip list in the form
+    # "192.168.1.2, 192.168.1.3"
+    ip = request.config.getoption('--jail_ip') or ''
+    ips = []
+    for check_ip in filter(bool, map(str.strip, ip.split(','))):
+        try:
+            ipaddress.IPv4Address(check_ip)
+        except ValueError:
+            pass
+        else:
+            ips.append(check_ip)
+
+    return ','.join(ips) or None
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -137,19 +137,7 @@ def zpool(request):
 @pytest.fixture
 def jail_ip(request):
     """Specify a jail ip to use."""
-    # For tests we only support ip list in the form
-    # "192.168.1.2, 192.168.1.3"
-    ip = request.config.getoption('--jail_ip') or ''
-    ips = []
-    for check_ip in filter(bool, map(str.strip, ip.split(','))):
-        try:
-            ipaddress.IPv4Address(check_ip)
-        except ValueError:
-            pass
-        else:
-            ips.append(check_ip)
-
-    return ','.join(ips) or None
+    return request.config.getoption('--jail_ip')
 
 
 @pytest.fixture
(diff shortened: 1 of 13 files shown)


tests/data_classes.py
```diff
@@ -521,14 +567,29 @@ def is_cloned(self):
 
     @property
     def ip(self):
+        try:
+            config = self.config
+            if config.get('dhcp', 'off') == 'on':
+                interface = 'epair0b'
+            else:
+                interface = config.get('ip4_addr', '').split('|')[0]
```
skarekrow: If they don't supply it in that format, this will not be a valid interface.

tests/functional_tests/0003_create_test.py
```diff
@@ -161,27 +156,46 @@ def test_09_test_dhcp_connectivity_in_jail(release, jail, invoke_cli):
     ip = jail.ip
     assert ip is not None and '0.0.0.0' not in ip
 
+    # Let's test if we can ping 8.8.8.8 from inside the jail
+
+    stdout, stderr = jail.run_command(['ping', '-c', '5', '8.8.8.8'])
```
skarekrow: Perhaps we should allow this to be configurable, there may be users who do not want Google to be aware of each test :D

tests/functional_tests/0003_create_test.py
```diff
@@ -161,27 +156,46 @@ def test_09_test_dhcp_connectivity_in_jail(release, jail, invoke_cli):
...
     try:
         with os.fdopen(fd, 'w') as tmp:
             tmp.write(json.dumps({
-                'pkgs': ['nginx']
+                'pkgs': install_pkgs
             }))
 
         invoke_cli(
-            ['create', '-r', release, '-n', 'pkg_jail', '-p', path]
+            ['create', '-r', release, '-n', pkg_jail, '-p', path, 'dhcp=on']
```
skarekrow: We should allow the pkg test to happen if they don't supply dhcp.

tests/functional_tests/0003_create_test.py
```diff
@@ -243,3 +257,22 @@ def test_14_create_rc_jail(release, jail, invoke_cli):
...
+
+
+@require_root
+@require_zpool
+@require_jail_ip
+def test_15_create_jail_with_assigned_ip(release, jail, invoke_cli, jail_ip):
+    jail = jail('assigned_ip_jail')
+
+    invoke_cli([
+        'create', '-r', release, '-n', jail.name, f'ip4_addr=vtnet0|{jail_ip}'
```
skarekrow: hardcoded vtnet0 :)

skarekrow (changes requested):

tests/functional_tests/0003_create_test.py
sonicaj: As in with a manual ip address ?

sonicaj: Networking is required for this test

tests/functional_tests/0003_create_test.py
sonicaj: Yeah 😓 As we want to run tests only, we can keep the user interaction as less as possible as the only intention is to see if the functionality is working as desired. However, maybe we can take input of interface and if not supplied default to `vtnet0` ? Thoughts ?

tests/functional_tests/0003_create_test.py
skarekrow: Yeah with a manual IP that they can specify. If they do, and not dhcp then they can't actually test this.

tests/functional_tests/0003_create_test.py
skarekrow: I think it's best to just drop vtnet0 entirely as iocage will handle that networking side of things with jail(8)

sonicaj: I will address this right away ;)

tests/conftest.py
```diff
@@ -146,7 +137,19 @@ def zpool(request):
 @pytest.fixture
 def jail_ip(request):
     """Specify a jail ip to use."""
-    return request.config.getoption('--jail_ip')
+    # For tests we only support ip list in the form
+    # "192.168.1.2, 192.168.1.3"
```
skarekrow: We don't support the user supplying an interface so it can properly alias?

tests/data_classes.py
```diff
@@ -131,7 +131,7 @@ def normalize_state(self):
 
     def normalize_ip4(self):
 
-        ip = self.ip4.split('|')
+        ip = str(self.ip4).split('|')
```
skarekrow, line 134: Is `self.ip4` not a str before this?

tests/data_classes.py
```diff
@@ -72,7 +72,7 @@ def __hash__(self):
     def normalize_name(self):
         # Logic copied from iocage to be consistent with the tests
         if self.name:
-            _sort = self.name.strip().rsplit('_', 1)
+            _sort = str(self.name).strip().rsplit('_', 1)
```
skarekrow, line 75: is `self.name` not a str before this?

tests/data_classes.py
```diff
@@ -177,7 +177,7 @@ def sort_flag(self, flag):
             value = (value,)
 
         # Default to sorting to name always
-        value += (self.name,)
+        value += (self.orig_name,)
```
skarekrow: Why are we using `orig_name`? Don't we want sorting to work on the renamed jails in the tests?

tests/data_classes.py
```diff
@@ -566,20 +565,36 @@ def is_cloned(self):
...
     @property
-    def ip(self):
+    def ips(self):
         try:
-            config = self.config
-            if config.get('dhcp', 'off') == 'on':
-                interface = 'epair0b'
-            else:
-                interface = config.get('ip4_addr', '').split('|')[0]
+            output = self.run_command(['ifconfig'])[0].split('\n')
```
skarekrow, line 585: Let's not parse output for stuff like this. ifconfig can change, we should use netifaces.

tests/data_classes.py
```diff
@@ -566,20 +565,36 @@ def is_cloned(self):
...
+        else:
+            skip_ips = ['127.0.0.1']
+            return [
+                ip for ip in ips if ip not in skip_ips
+            ]
 
-        except subprocess.CalledProcessError:
-            pass
+    @property
+    def ip(self):
```
skarekrow: Perhaps rename to first_ip? Or we can allow specifying an index

tests/functional_tests/0002_fetch_test.py
```diff
@@ -31,10 +31,10 @@
 @require_root
 @require_zpool
 def test_fetch(
-        release, server, user, password, auth,
-        root_dir, http, _file, noupdate, invoke_cli
+    release, server, user, password, auth,
+    root_dir, http, _file, noupdate, invoke_cli
 ):
-
+    return
```
skarekrow: debug?

tests/functional_tests/0097_upgrade_test.py
```diff
@@ -65,11 +66,16 @@ def test_01_upgrade_jail(
 
     jail = jail('upgrade_test')
     # Let's create a jail now of version releases.index(release) - 1
+    if (jail_ip and dhcp) or jail_ip:
```
skarekrow: We can just simplify this to just `if jail_ip`, as we don't care if they supplied dhcp or not at that point.

skarekrow (changes requested):

tests/functional_tests/0096_update_test.py
```diff
@@ -0,0 +1,58 @@
...
+@require_networking
+@require_root
+@require_zpool
+def test_01_upgrade_jail(
+    invoke_cli, jail, release, dhcp, jail_ip
+):
+
+    jail = jail('update_test')
+    # Let's create a jail now
+    if (jail_ip and dhcp) or jail_ip:
```
skarekrow: We can just use `if jail_ip` here again

skarekrow (changes requested):

tests/data_classes.py
sonicaj: Precautionary measures :P

skarekrow (dismissed):

skarekrow (approved):

PR #767, opened 2018-12-30
Add allow.vmm jail parameter

jamesog:
FreeBSD 12.0-RELEASE has a new jail(8) parameter "allow.vmm" to allow
jails to access vmm(4).

This adds a new allow_vmm config value to iocage to support this and
bumps the config version to 17.

- [x] Explain the feature

### DIFF
--- a/iocage.8
+++ b/iocage.8
@@ -1630,6 +1630,21 @@ Default: 0
 .Pp
 Source:
 .Xr mlock 2
+.It Pf allow_vmm= Op 0 | 1
+Allow access to
+.Xr vmm 4
+inside the jail. The
+.Xr vmm 4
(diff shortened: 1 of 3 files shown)


skarekrow (changes requested):
Thanks for this! A few things:
1 - You'll want to add this to the section around line 263 of ioc_start since it requires 12
2 - You'll want to add this to the default_props dictionary at line 675 so the default configuration has this.

jamesog:
Thanks @skarekrow! I'll get those changes made shortly. I actually had the second one done on the machine I was testing on but it got lost in copy-pasta 😳.

skarekrow:
Thanks again!

Issue #758, opened 2018-12-25
Don't remove aliases on shared networking


pos42:
Hi

I have host on 12.0 release p1 AMD64 (upgrade from 11.2 a week ago) and my jails on 11.2 release p5. I use shared IP and all is working with iocage 0.9.10 except a few things I was waiting for in 1.0 (like jail upgrades...).

So I just upgraded to 1.0 that just came in ports. I then rebuilt all ports.

Now…
At a host reboot the system by itself change my IPv4 address from a /24 to a /32 mask. The host therefor stopped working on IPv4. In /etc/rc.conf the ifconfig_vtnet0 has a 255.255.255.0 mask and it has always worked…. For the host to work I have to manually run ifconfig and set the /24 mask back and also manually run route add default….

And when stopping a jail the whole IPv4 address is wiped from the host.

1.0 is not working as expected.

I reverted to 0.9.10 ad this was working again. Is this a known problem, or am I missing something?

/Peo

phyber:
This appears to be because iocage doesn't trust FreeBSD or the system adminsitrator to manage IP addresses properly, so iocage /always/ removes IP addresses when the jail is stopped. Additionally, when starting a jail it does some extra work to try and see which interface it should prepend onto the `ip[4|6].addr` if none was given, which would result in FreeBSD removing the address on stop if iocage wasn't already doing it. Both of these things seem very wrong.

If FreeBSD was allowed to manage this, everything should "just work".

  - `ip[4|6].addr` in the form of `interface|ip-address[/cidr]` would have FreeBSD configure the `ip-address` on the given `interface` at jail start, and remove it on stop.
  - The simple `ip-address[/cidr]` form would have FreeBSD trust the system administrator to know what they're doing with their IP addressing, so FreeBSD would neither add nor remove addresses from interfaces.

I believe iocage should either:

  - Just let FreeBSD manage things, as documented in `jail(8)`, or
  - Introduce an option so that people can request regular, `jail(8)` style management for their networking.

My personal preference is the former, and I am planning to hack at iocage to make it work as expected. I understand upstream will have concerns over backwards compatibility, so I also plan to try to add an option to make iocage behave the way it currently does (but default it to off).

I am also aware that there may be good reasons for iocage currently behaving the way it does, however, I didn't notice any comments around the code saying why it does this, nor did I see anything on the [readthedocs](https://iocage.readthedocs.io/en/latest/networking.html) page.

pos42:
[...]

PR #770, opened 2019-01-01
Leverage jail(8) for shared IP networking behaviour

phyber:
As discussed in #758, this PR changes the behaviour of `iocage` to let `jail(8)` take on the heavy lifting of managing shared IP jails.

The main change to enable this behaviour is that `iocage` will now write out a `jail.conf` for each jail it manages under `/var/run`, it will then use these configs when starting and stopping the jails.
This allows users to (ip4 given as examples, but it works the same for ip6):
  - Have jailed shared with the host, managed via `/etc/rc.conf`, `ip4.addr=192.168.1.1`.
  - Have jails share IPs with each other, using the above.
  - Have `iocage` (really `jail(8)`) manage the addition and removal of aliases on jail start/stop, `ip4.addr=lo1|127.0.1.1`, etc.

The old behaviour of `iocage` trying to work out:
  - On start: If it should add the "primary" network interface to an `ip4_addr` that was given simply as `1.2.3.4`.
  - On stop: Which `ip4_addr` aliases it should remove via `ifconfig`.

have been completely removed. It would be possible to support both the old and new behaviour, but it gets quite complicated, and I'm not sure there's a good reason to keep them when FreeBSD can manage it via `jail(8)`.

Another small side effect of these changes was that we fixed a small bug where `iocage` could orphan addresses on interfaces if the `ip4_addr` was changed while a jail was running and the jail was then restarted. Vim backup files `*~` were also added to `.gitignore`.

I have not yet squashed this PR, as if changes are requested my history might be useful, but I will squash it if/when the PR is accepted.
I have also not yet fixed up the documentation, pending review of the work so far.

While testing this I've been managing 6-7 jails with varying combinations of address setups successfully, however, I only use shared IP behaviours so I have not tested `vnet` based jails. I believe they should be working as they were before, but someone else may want to test that.

- [x] Explain the feature
- [ ] Documentation updated
- [ ] Squashed

### DIFF
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -392,6 +392,7 @@ class IOCStart(object):
         except OSError as err:
             msg = f"Error while writing jail config {err.filename}: " \
                    + "{err.strerror}"
+
             iocage_lib.ioc_common.logit({
                 "level": "EXCEPTION",
                 "message": msg
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -275,13 +275,7 @@ class IOCStart(object):
             net = []
 
             if ip4_addr != "none":
-                ip4_addrs = []
-
-                for _ip4_addr in ip4_addr.split(","):
-                    ip4_addrs.append(f"{_ip4_addr}")
-
-                ip4_addrs_str = ",".join(ip4_addrs)
-                net.append(f"ip4.addr={ip4_addrs_str}")
+                net.append(f"ip4.addr={ip4_addr}")
 
             if ip6_addr != "none":
                 net.append(f"ip6.addr={ip6_addr}")
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -369,8 +369,8 @@ class IOCStart(object):
             f"allow.mount.zfs={allow_mount_zfs}"
         ]
 
-        start_cmd = [x for x in ["jail", "-c"] + net + [
-            x for x in parameters if '1' in x] + [
+        start_parameters = [x for x in [
+            net + [x for x in parameters if '1' in x] + [
                 f'name=ioc-{self.uuid}',
                 f'host.domainname={host_domainname}',
                 f'host.hostname={host_hostname}',
@@ -392,7 +392,10 @@ class IOCStart(object):
                 f'exec.consolelog={self.iocroot}/log/ioc-'
                 f'{self.uuid}-console.log',
                 'persist'
+            ]
         ] if x != '']
+
+        start_cmd = ["jail", "-c"] + start_parameters
 
         start_env = {
             **os.environ,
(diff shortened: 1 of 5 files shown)


phyber:
Will take care of the travis issues tomorrow. 👍

iocage_lib/ioc_start.py
```diff
@@ -275,18 +275,13 @@ def __start_jail__(self):
...
-                            def_iface = gws["default"][netifaces.AF_INET][1]
-                            _ip4_addr = f'{def_iface}|{_ip4_addr}'
-                        except KeyError:
-                            # Best effort for default interface
-                            pass
+                    ip4_addrs.append(f"{_ip4_addr}")
 
-                    net.append(f"ip4.addr={_ip4_addr}")
+                ip4_addrs_str = ",".join(ip4_addrs)
+                net.append(f"ip4.addr={ip4_addrs_str}")
```
skarekrow: I think these lines can be squashed down to a single line, as ip4_addrs isn't used anywhere else

```
net.append(f'ip4.addr={ip4_addr}')
```

skarekrow (changes requested):
Looks good! Only potential downside right now is that we should remove these sentinel files on stop, as they're just rewritten anyways on start. We should also check for and remove them during destroy/clean. These may pile up quite a bit.

iocage_lib/ioc_start.py
phyber: Will put that in with the next changes.

phyber:
Fair comment, I did wonder about deleting them on stop, and this is a good reason to do so.

skarekrow (dismissed):

phyber:
I _think_ the changes to `ioc_destroy.py` means that the jail conf files should also be cleaned via the appropriate things in `ioc_clean.py` as they go via the `destroy_jail` function.

skarekrow:
> I _think_ the changes to `ioc_destroy.py` means that the jail conf files should also be cleaned via the appropriate things in `ioc_clean.py` as they go via the `destroy_jail` function.

Yep that should do it. We should perhaps catch exceptions with the destruction process and pass since it's just a sentinel file, and not something that should hold up destroy/clean

skarekrow (approved):

phyber:
`OSError`s on `unlink` in both stop and destroy are now wrapped to `pass` on failure. I wasn't sure if you wanted the `su.Popen` call to `jail` wrapped in the destroy too, so I've left that alone for the moment.

skarekrow:
Nope I think those should still be viable errors to see. Since we actually want the jail process removed. Thanks again for this PR!

phyber:
If we're happy here, I'll perform the final squash. It doesn't look like anything in the existing documentation really needs changing.

phyber:
Or, I guess github has that Squash Merge button these days :)

skarekrow:
As long as you are done with this I can actually squash and commit it as is. I don't think any documentation should need changing, since this is all (much needed) plumbing.

phyber:
Yup, I believe I'm done if you're happy with it as is.

skarekrow:
:+1: Thanks!

PR #774, opened 2019-01-04
Fix various Tracebacks for non-root users

jsegaert:

- [X] Explain the feature

This fixes a few Tracebacks that occur when running iocage as a non-privileged user, e.g.
- iocage list, the first time after bumping the config version
- iocage get -P, to get plugin specific properties either wants to start the jail or run jexec if the jails is already running


skarekrow (approved):
Thanks!

PR #775, opened 2019-01-04
Improve exception handling for listing

sonicaj:
This commit improves exception handling for a case when we tried to determine a jail's ip address and failed to parse the output.
Ticket: #65931

### DIFF
--- a/iocage_lib/ioc_list.py
+++ b/iocage_lib/ioc_list.py
@@ -228,6 +228,7 @@ class IOCList(object):
                 ip6 = "-"
 
             # Will be set already by a corrupt jail
+            status = False
             if state != 'CORRUPT':
                 status, jid = self.list_get_jid(uuid_full)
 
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_list.py
```diff
@@ -268,8 +269,20 @@ def list_all(self, jails):
                     short_ip4 += "(Network Issue)"
                     full_ip4 = f"DHCP - Network Issue:{e}"
                 else:
-                    full_ip4 = f"{interface}|" \
-                        f"{out.splitlines()[2].split()[1].decode()}"
+                    try:
+                        full_ip4 = f'{interface}|' \
+                            f'{out.splitlines()[2].split()[1].decode()}'
+                    except IndexError as e:
+                        short_ip4 = '(Network Issue)'
```
skarekrow: Let’s use the same DHCP - Failed Parsing message without the rest here.

skarekrow (changes requested):

iocage_lib/ioc_list.py
sonicaj: Should we log it still ? I think we should perhaps in a case where we have out but fail to parse the output of the command ? Thoughts ?

skarekrow: I think in this case the list is essentially a log. If we log it, that will print and clutter things up. Until we have debug logs and a verbose mode anyways.

skarekrow (approved):

sonicaj:
backport

PR #776, opened 2019-01-04
Add ignore exceptions flag

sonicaj:
This commit introduces a flag for suppressing exceptions when starting/stopping jails. This can be useful when we have jails which cannot really start and we don't want the system to exit in this case.
Ticket: #65391

### DIFF
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -70,7 +70,7 @@ class IOCStart(object):
             self.exec_fib = self.conf["exec_fib"]
             try:
                 self.__start_jail__()
-            except SystemExit as e:
+            except (Exception, SystemExit) as e:
                 if not suppress_exception:
                     raise e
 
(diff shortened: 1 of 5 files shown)


iocage_lib/ioc_start.py
```diff
@@ -66,7 +68,11 @@ def __init__(self, uuid, path, silent=False, callback=None,
                                                    silent=True).json_set_value
 
             self.exec_fib = self.conf["exec_fib"]
-            self.__start_jail__()
+            try:
+                self.__start_jail__()
+            except SystemExit as e:
```
skarekrow: That does mean that if iocage is ran noninteractively (freenas mw) then the exceptions will never be suppressed. Is that the desired behavior?

skarekrow (changes requested):
Perhaps we make ignore exception a self property so we can minimize the number of changed calls? We may want this on other things down the line too.

skarekrow (approved):

PR #777, opened 2019-01-04
Check Release exists

sonicaj:
This commit makes sure we check the release exists before destroying it. If it doesn't, we raise an appropriate exception outlining that it does not exist. It also introduces a new class Resource which we can use to manipulate resources in iocage. However much thought is still needed on how we should proceed with it in the future.
Ticket: #63810

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -223,7 +223,7 @@ class Release(Resource):
         # the release, let's normalize it
         # a name can't contain "/". If it's a path, we can make a split and
         # use the last name
-        super().__init__(name if '/' not in name else name.split('/')[1])
+        super().__init__(name if '/' not in name else name.rsplit('/', 1)[1])
 
     @property
     def path(self):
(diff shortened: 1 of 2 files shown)


iocage_lib/ioc_json.py
```diff
@@ -153,11 +181,57 @@ def zfs_get_dataset_name(self, name):
...
+        return os.path.exists(self.path)
+
+
+class Release(Resource):
+    def __init__(self, name):
+        # We can expect the name to be either a full path or just the name of
+        # the release, let's normalize it
+        # a name can't contain "/". If it's a path, we can make a split and
+        # use the last name
+        super().__init__(name if '/' not in name else name.split('/')[1])
```
skarekrow: We should rsplit on this

iocage_lib/ioc_json.py
```diff
@@ -110,6 +110,34 @@ class IOCZFS(object):
     def __init__(self):
         self.zfs = libzfs.ZFS(history=True, history_prefix="<iocage>")
 
+    @property
+    def iocroot_path(self):
+        # For now we assume iocroot is set, however moving on we need to think
+        # how best to structure this
+        # We should give thought to how we handle
+        for pool in self.pools:
```
skarekrow, line 118: This only returns the first instance, which on a misconfigured system could be more than 1

skarekrow (changes requested):

skarekrow (approved):

PR #782, opened 2019-01-05
tkt-67800: Jail Configuration Class

sonicaj:
This PR introduces the following changes:
1) Introduce a class to better manage jail.conf files
2) Make sure we don't perform post stop operations if jail process fails to exit
3) Bug fix for a case where we ran a few exec* operations twice

### DIFF
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -38,6 +38,7 @@ import re
 import shlex
 import glob
 
+import iocage_lib.ioc_exceptions
 import iocage_lib.ioc_exec
 
 
@@ -787,12 +788,15 @@ def runscript(script):
     else:
         return False, 'Script is not executable!'
 
-    with iocage_lib.ioc_exec.IOCExec(
-        script, None, None, unjailed=True, decode=True
-    ) as _exec:
-        success, error = list(_exec)[0]
-
-    return success.rstrip('\n'), error.rstrip('\n')
+    try:
+        with iocage_lib.ioc_exec.IOCExec(
+            script, None, uuid='', unjailed=True, decode=True
+        ) as _exec:
+            success, error = list(_exec)[0]
+    except iocage_lib.ioc_exceptions.CommandFailed as e:
+        return '', f'Script returned non-zero status: {e}'
+    else:
+        return success.rstrip('\n'), error.rstrip('\n')
 
 
 def match_to_dir(iocroot, uuid, old_uuid=None):
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -38,6 +38,8 @@ import re
 import shlex
 import glob
 
+import iocage_lib.ioc_exec
+
 
 def callback(_log, callback_exception):
     """Helper to call the appropriate logging level"""
@@ -573,22 +575,6 @@ def checkoutput(*args, **kwargs):
     return out
 
 
-def safe_checkoutput(*args, **kwargs):
-    """
-    Returns a tuple, first value stating the success of the call returning
-    True on success and False otherwise. The second value the stderr/stdout
-    if any
-    """
-    success = False
-    try:
-        out = checkoutput(*args, **kwargs)
-        success = True
-    except su.CalledProcessError as e:
-        out = e.output.decode('utf-8')
-
-    return success, out
-
-
 def set_rcconf(jail_path, key, value):
     conf_file = f"{jail_path}/root/etc/rc.conf"
 
@@ -787,8 +773,8 @@ def generate_devfs_ruleset(conf, paths=None, includes=None, callback=None,
 
 def runscript(script):
     """
-    Runs the script provided and return a tuple with first value determining
-    success of the script and last showing stderr/stdout if any
+    Runs the script provided and return a tuple with first value showing
+    stdout and last showing stderr
     """
     script = shlex.split(script)
 
@@ -801,9 +787,12 @@ def runscript(script):
     else:
         return False, 'Script is not executable!'
 
-    success, output = safe_checkoutput(script, stderr=su.STDOUT)
+    with iocage_lib.ioc_exec.IOCExec(
+        script, None, None, unjailed=True, decode=True
+    ) as _exec:
+        success, error = list(_exec)[0]
 
-    return success, output.rstrip('\n')
+    return success.rstrip('\n'), error.rstrip('\n')
 
 
 def match_to_dir(iocroot, uuid, old_uuid=None):
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -802,6 +802,7 @@ def runscript(script):
         return False, 'Script is not executable!'
 
     success, output = safe_checkoutput(script, stderr=su.STDOUT)
+
     return success, output.rstrip('\n')
 
 
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -566,11 +566,27 @@ def checkoutput(*args, **kwargs):
     try:
         out = su.check_output(*args, **kwargs)
 
-        out = out.decode("utf-8")
+        out = out.decode('utf-8')
     except su.CalledProcessError:
         raise
 
     return out
+
+
+def safe_checkoutput(*args, **kwargs):
+    """
+    Returns a tuple, first value stating the success of the call returning
+    True on success and False otherwise. The second value the stderr/stdout
+    if any
+    """
+    success = False
+    try:
+        out = checkoutput(*args, **kwargs)
+        success = True
+    except su.CalledProcessError as e:
+        out = e.output.decode('utf-8')
+
+    return success, out
 
 
 def set_rcconf(jail_path, key, value):
@@ -771,29 +787,26 @@ def generate_devfs_ruleset(conf, paths=None, includes=None, callback=None,
 
 def runscript(script):
     """
-    Runs the users provided script, otherwise returns a tuple with
-    True/False and the error.
+    Runs the script provided and return a tuple with first value determining
+    success of the script and last showing stderr/stdout if any
     """
     script = shlex.split(script)
 
     if len(script) > 1:
         # We may be getting ';', '&&' and so forth. Adding the shell for
         # safety.
-        script = ["/bin/sh", "-c", " ".join(script)]
+        script = ['/bin/sh', '-c', ' '.join(script)]
     elif os.access(script[0], os.X_OK):
         script = script[0]
     else:
-        return True, "Script is not executable!"
+        return False, 'Script is not executable!'
 
     try:
         out = checkoutput(script, stderr=su.STDOUT)
     except su.CalledProcessError as err:
-        return False, err.output.decode().rstrip("\n")
+        return False, err.output.decode().rstrip('\n')
 
-    if out:
-        return True, out.rstrip("\n")
-
-    return True, None
+    return True, out.rstrip('\n')
 
 
 def match_to_dir(iocroot, uuid, old_uuid=None):
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -801,12 +801,8 @@ def runscript(script):
     else:
         return False, 'Script is not executable!'
 
-    try:
-        out = checkoutput(script, stderr=su.STDOUT)
-    except su.CalledProcessError as err:
-        return False, err.output.decode().rstrip('\n')
-
-    return True, out.rstrip('\n')
+    success, output = safe_checkoutput(script, stderr=su.STDOUT)
+    return success, output.rstrip('\n')
 
 
 def match_to_dir(iocroot, uuid, old_uuid=None):
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -789,14 +789,13 @@ def runscript(script):
         return False, 'Script is not executable!'
 
     try:
-        with iocage_lib.ioc_exec.IOCExec(
-            script, None, uuid='', unjailed=True, decode=True
-        ) as _exec:
-            success, error = list(_exec)[0]
+        output = iocage_lib.ioc_exec.SilentExec(
+            script, None, unjailed=True, decode=True
+        )
     except iocage_lib.ioc_exceptions.CommandFailed as e:
         return '', f'Script returned non-zero status: {e}'
     else:
-        return success.rstrip('\n'), error.rstrip('\n')
+        return output.stdout.rstrip('\n'), output.stderr.rstrip('\n')
 
 
 def match_to_dir(iocroot, uuid, old_uuid=None):
--- a/iocage_lib/ioc_exec.py
+++ b/iocage_lib/ioc_exec.py
@@ -38,20 +38,23 @@ import collections
 
 class IOCExec(object):
     """Run jexec with a user inside the specified jail."""
-    def __init__(self,
-                 command,
-                 path,
-                 # None is special for RELEASE updating to work around
-                 # freebsd-update weirdness
-                 uuid='',
-                 host_user="root",
-                 jail_user=None,
-                 plugin=False,
-                 unjailed=False,
-                 skip=False,
-                 stdin_bytestring=None,
-                 su_env=None,
-                 callback=None):
+    def __init__(
+        self,
+        command,
+        path,
+        # None is special for RELEASE updating to work around
+        # freebsd-update weirdness
+        uuid='',
+        host_user='root',
+        jail_user=None,
+        plugin=False,
+        unjailed=False,
+        skip=False,
+        stdin_bytestring=None,
+        su_env=None,
+        decode=False,
+        callback=None
+    ):
         self.command = command
         self.uuid = uuid.replace(".", "_") if uuid is not None else uuid
         self.path = path
@@ -61,6 +64,7 @@ class IOCExec(object):
         self.unjailed = unjailed
         self.skip = skip
         self.stdin_bytestring = stdin_bytestring
+        self.decode = decode
         self.stdin = su.PIPE if self.stdin_bytestring is not None else None
 
         path = '/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:'\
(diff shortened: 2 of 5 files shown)


iocage_lib/ioc_stop.py
```diff
@@ -85,39 +85,7 @@ def __stop_jail__(self):
...
-                    msg = f"  + Running prestop FAILED"
-
-                iocage_lib.ioc_common.logit({
-                    "level": "ERROR",
-                    "message": msg
-                },
-                    _callback=self.callback,
-                    silent=self.silent)
-
+            # TODO: Perhaps have jail(8) handle this ?
```
sonicaj: Actually @skarekrow I think we definitely should now. As prestop would be handled by jail(8), it would not make sense to have `stop` executed before that

iocage_lib/ioc_stop.py
```diff
@@ -285,32 +264,32 @@ def __stop_jail__(self):
...
+            cmd.extend(["-f", f"{jail_conf_file}"])
 
-            cmd.extend(["-r", f"ioc-{self.uuid}"])
+        cmd.extend(["-r", f"ioc-{self.uuid}"])
 
-            stop = su.check_call(cmd, stderr=su.PIPE)
-        except su.CalledProcessError as err:
-            stop = err.returncode
+        stop = su.Popen(cmd, stdout=su.PIPE, stderr=su.PIPE)
+        success, stop_err = stop.communicate()
```
skarekrow: Since we don't use `success`, let's just make it `_`

iocage_lib/ioc_stop.py
```diff
@@ -285,32 +264,32 @@ def __stop_jail__(self):
...
+        # will be written on the next jail start/restart.
+        jail_conf_file = Path(f"/var/run/jail.ioc-{self.uuid}.conf")
 
-            if jail_conf_file.is_file():
-                cmd.extend(["-f", f"{jail_conf_file}"])
+        if jail_conf_file.is_file():
+            cmd.extend(["-f", f"{jail_conf_file}"])
 
-            cmd.extend(["-r", f"ioc-{self.uuid}"])
+        cmd.extend(["-r", f"ioc-{self.uuid}"])
```
skarekrow: Let's stick with single quotes like the rest of the new additions :D *OCD*

iocage_lib/ioc_stop.py
```diff
@@ -195,6 +163,17 @@ def __stop_jail__(self):
                                 "{}".format(
                                     err.output.decode("utf-8").rstrip()))
 
+        else:
```
skarekrow, line 196: Else? This won't get ran if they have jailed zfs datasets then

iocage_lib/ioc_stop.py
skarekrow: As we discussed offline, it's prudent atm to handle all exec.* iocage side until we have some other breaking things for 1.2

iocage_lib/ioc_start.py
```diff
@@ -392,18 +392,10 @@ def __start_jail__(self):
...
-            msg = f"Error while writing jail config {err.filename}: " \
-                  + "{err.strerror}"
-
-            iocage_lib.ioc_common.logit({
-                "level": "EXCEPTION",
-                "message": msg
-            },
-                _callback=self.callback,
-                silent=self.silent)
+        jail = iocage_lib.ioc_json.JailConfiguration(
```
skarekrow: My concern is we now don't handle any exceptions in the class itself.

iocage_lib/ioc_json.py
```diff
@@ -45,6 +45,121 @@
...
+
+        # Let's treat ip4.addr and ip6.addr differently keeping compatibility
+        # with current jail.conf files we have introduced
+        ip4 = []
+        ip6 = []
+        for data in content:
+            if '=' in data:
+                k, v = data.split('=', 1)
+                k = k.strip()
+                v = v.replace(';', '').strip()
```
skarekrow, line 96: Can we use some line breaks for some logic separation in general :P

iocage_lib/ioc_json.py
```diff
@@ -45,6 +45,121 @@
...
+                    self.read_data[k] = v
+            else:
+                # None is special for self.data value as it indicates that
+                # the key is a boolean one, we will write a value of empty
+                # string as a key,value pair and it will NOT be treated as
+                # a boolean value
+                self.read_data[data.replace(';', '').strip()] = None
+
+        if ip4:
+            self.read_data['ip4.addr'] = ', '.join(ip4)
```
skarekrow: I don't believe we need the whitespace

iocage_lib/ioc_json.py
```diff
@@ -45,6 +45,121 @@
...
+        if ip6:
+            self.read_data['ip6.addr'] = ', '.join(ip6)
```
skarekrow: Same here

skarekrow (changes requested):

iocage_lib/ioc_common.py
```diff
@@ -566,13 +566,29 @@ def checkoutput(*args, **kwargs):
...
 
-        out = out.decode("utf-8")
+        out = out.decode('utf-8')
     except su.CalledProcessError:
         raise
 
     return out
 
 
+def safe_checkoutput(*args, **kwargs):
```
skarekrow: Perhaps we just replace our checkoutput method with this?

iocage_lib/ioc_start.py
```diff
@@ -525,30 +520,66 @@ def __start_jail__(self):
         self.start_generate_resolv()
         self.start_copy_localtime()
         # This needs to be a list.
-        exec_start = self.conf["exec_start"].split()
+        exec_start = self.conf['exec_start'].split()
+
+        with open(
+            os.path.join(self.iocroot, 'log', f'{self.uuid}-console.log'), 'a'
```
skarekrow: Let's just use an fsting here instead of the `.format()` or `os` call

iocage_lib/ioc_start.py
```diff
@@ -525,30 +520,66 @@ def __start_jail__(self):
...
-                "message": msg
+                'level': 'EXCEPTION',
+                'message': msg
             },
                 _callback=self.callback,
-                silent=self.silent)
+                silent=self.silent
+            )
+
+            iocage_lib.ioc_stop.IOCStop(
```
skarekrow: This needs to run before the logit, as otherwise the program will be dead.

iocage_lib/ioc_start.py
```diff
@@ -525,30 +520,66 @@ def __start_jail__(self):
...
-            services = su.check_call(["setfib", self.exec_fib, "jexec",
-                                      f"ioc-{self.uuid}"] + exec_start,
-                                     stdout=f, stderr=su.PIPE)
+            f.write(output)
 
-        if services:
-            msg = "  + Starting services FAILED"
+        if not success:
+            msg = f'  + Starting services FAILED\nERROR:\n{output}\n\n' \
+                'exec_start failed to execute.\nJail failed to start'
```
skarekrow: Perhaps it's clearer if we say: 
```
            msg = f'  + Starting services FAILED\nERROR:\n{output}\n\n' \
                       f'Refusing to start {self.uuid}: exec_start failedt'
```

iocage_lib/ioc_start.py
```diff
@@ -525,30 +520,66 @@ def __start_jail__(self):
...
+            poststart_success, poststop_output = iocage_lib.ioc_common.runscript(
+                exec_poststart
+            )
+
+            if not poststart_success:
+                iocage_lib.ioc_common.logit({
+                    'level': 'EXCEPTION',
+                    'message': '  + Executing exec_poststart FAILED\n'
+                    f'ERROR:\n{poststop_output}\n\nJail '
+                    f'failed to start'
```
skarekrow: Same format as above for `exec_start`

iocage_lib/ioc_start.py
```diff
@@ -525,30 +520,66 @@ def __start_jail__(self):
...
-                silent=self.silent)
+                silent=self.silent
+            )
+
+            # Running exec_poststart now
+            poststart_success, poststop_output = iocage_lib.ioc_common.runscript(
+                exec_poststart
+            )
+
+            if not poststart_success:
```
skarekrow: We're missing the stop call after this

iocage_lib/ioc_stop.py
```diff
@@ -84,33 +84,60 @@ def __stop_jail__(self):
             _callback=self.callback,
             silent=self.silent)
 
+        failed_message = 'Please use force flag to force stop jail'
```
skarekrow: `--force` instead of force flag

iocage_lib/ioc_stop.py
```diff
@@ -307,46 +335,41 @@ def __stop_jail__(self):
...
-                _callback=self.callback,
-                silent=self.silent)
-        elif poststop:
-            msg = "  + Running poststop OK"
+        if not poststop_success:
+            # This is the only exec case where we won't raise an exception
+            # as jail has already stopped
+            msg = f'  + Running poststop FAILED\n{poststop_output}\n\n' \
+                'Jail has been stopped but there might be leftovers ' \
+                'from poststop failure'
```
skarekrow: `exec_poststop`

iocage_lib/ioc_stop.py
```diff
@@ -307,46 +335,41 @@ def __stop_jail__(self):
...
-            },
-                _callback=self.callback,
-                silent=self.silent)
-        elif poststop:
-            msg = "  + Running poststop OK"
+        if not poststop_success:
+            # This is the only exec case where we won't raise an exception
+            # as jail has already stopped
+            msg = f'  + Running poststop FAILED\n{poststop_output}\n\n' \
+                'Jail has been stopped but there might be leftovers ' \
```
skarekrow: `may be` instead of `might be`

skarekrow (changes requested):

iocage_lib/ioc_common.py
```diff
@@ -801,12 +801,8 @@ def runscript(script):
...
         return False, 'Script is not executable!'
 
-    try:
-        out = checkoutput(script, stderr=su.STDOUT)
-    except su.CalledProcessError as err:
-        return False, err.output.decode().rstrip('\n')
-
-    return True, out.rstrip('\n')
+    success, output = safe_checkoutput(script, stderr=su.STDOUT)
+    return success, output.rstrip('\n')
```
skarekrow: line break :P

skarekrow (changes requested):

iocage_lib/ioc_start.py
sonicaj: Good catch ;)

iocage_lib/ioc_start.py
```diff
@@ -564,20 +565,21 @@ def __start_jail__(self):
...
+                iocage_lib.ioc_stop.IOCStop(
+                    self.uuid, self.path, force=True, silent=True
+                )
+
                 iocage_lib.ioc_common.logit({
                     'level': 'EXCEPTION',
-                    'message': '  + Executing exec_poststart FAILED\n'
-                    f'ERROR:\n{poststop_output}\n\nJail '
-                    f'failed to start'
+                    'message': '  + Running exec_poststart FAILED\n'
```
skarekrow: Should still be executing, as that is what we're doing. It may point a user to the fact that their script is not executable as well via language

skarekrow (changes requested):

iocage_lib/ioc_exec.py
```diff
@@ -38,18 +38,21 @@
...
-                 plugin=False,
-                 unjailed=False,
-                 skip=False,
-                 stdin_bytestring=None,
-                 su_env=None,
-                 callback=None):
+    def __init__(
+            self,
+            command,
+            uuid,
```
skarekrow: We can make this optional as None

iocage_lib/ioc_exec.py
```diff
@@ -38,18 +38,21 @@
...
+            path,
+            host_user="root",
+            jail_user=None,
+            plugin=False,
+            unjailed=False,
+            skip=False,
+            stdin_bytestring=None,
+            su_env=None,
+            callback=None,
+            decode=False
```
skarekrow: I like the new additions above callback usually, just a style niggle :P

iocage_lib/ioc_exec.py
```diff
@@ -229,7 +233,12 @@ def exec_jail(self):
                     rtrn_stderr = self.proc.stderr.read()
                     stderr_queue.append(rtrn_stderr)
 
-                yield rtrn_stdout, rtrn_stderr
+                if not self.decode:
+                    yield rtrn_stdout, rtrn_stderr
+                else:
+                    yield rtrn_stdout.decode('utf-8'), rtrn_stderr.decode(
```
skarekrow: Just `.decode()` is actually faster and is the same as `utf-8`

iocage_lib/ioc_exec.py
```diff
@@ -38,18 +38,21 @@
...
-                 unjailed=False,
-                 skip=False,
-                 stdin_bytestring=None,
-                 su_env=None,
-                 callback=None):
+    def __init__(
+            self,
+            command,
+            uuid,
+            path,
```
skarekrow: We can make this optional as well, and then do a sanity check if uuid is supplied that this also must be

iocage_lib/ioc_common.py
```diff
@@ -801,9 +787,12 @@ def runscript(script):
     else:
         return False, 'Script is not executable!'
 
-    success, output = safe_checkoutput(script, stderr=su.STDOUT)
+    with iocage_lib.ioc_exec.IOCExec(
+        script, None, None, unjailed=True, decode=True
+    ) as _exec:
+        success, error = list(_exec)[0]
```
skarekrow: Perhaps we extend SilentExec to actually return stdout, stderr instead of just passing and using a list on the generator here?

iocage_lib/ioc_start.py
```diff
@@ -525,20 +526,21 @@ def __start_jail__(self):
...
             f'{self.iocroot}/log/{self.uuid}-console.log', 'a'
         ) as f:
-            success, output = iocage_lib.ioc_common.safe_checkoutput(
+            with iocage_lib.ioc_exec.IOCExec(
                 ['setfib', self.exec_fib, 'jexec', f'ioc-{self.uuid}']
-                + exec_start, stderr=su.STDOUT
-            )
+                + exec_start, None, None, unjailed=True, decode=True
+            ) as _exec:
+                success, error = list(_exec)[0]
```
skarekrow: Same here

iocage_lib/ioc_stop.py
```diff
@@ -112,16 +113,17 @@ def __stop_jail__(self):
...
             exec_stop = self.conf['exec_stop'].split()
             with open(f'{self.iocroot}/log/{self.uuid}-console.log', 'a') as f:
-                success, output = iocage_lib.ioc_common.safe_checkoutput(
+                with iocage_lib.ioc_exec.IOCExec(
                     ['setfib', exec_fib, 'jexec', f'ioc-{self.uuid}']
-                    + exec_stop, stderr=su.STDOUT
-                )
+                    + exec_stop, None, None, unjailed=True, decode=True
+                ) as _exec:
+                    success, error = list(_exec)[0]
```
skarekrow: and here :D

iocage_lib/ioc_stop.py
```diff
@@ -335,14 +337,14 @@ def __stop_jail__(self):
...
+        poststop_success, poststop_error = iocage_lib.ioc_common.runscript(
             self.conf['exec_poststop']
         )
 
-        if not poststop_success:
+        if poststop_error:
             # This is the only exec case where we won't raise an exception
             # as jail has already stopped
-            msg = f'  + Running poststop FAILED\n{poststop_output}\n\n' \
+            msg = f'  + Running poststop FAILED\n{poststop_error}\n\n' \
```
skarekrow: We should probably say Executing on all the exec* stuff

skarekrow (changes requested):

iocage_lib/ioc_common.py
```diff
@@ -787,12 +788,15 @@ def runscript(script):
...
 
-    with iocage_lib.ioc_exec.IOCExec(
-        script, None, None, unjailed=True, decode=True
-    ) as _exec:
-        success, error = list(_exec)[0]
-
-    return success.rstrip('\n'), error.rstrip('\n')
+    try:
+        with iocage_lib.ioc_exec.IOCExec(
+            script, None, uuid='', unjailed=True, decode=True
```
skarekrow: `uuid` is already `''` by default :P

iocage_lib/ioc_start.py
```diff
@@ -537,7 +537,7 @@ def __start_jail__(self):
         ) as f:
             with iocage_lib.ioc_exec.IOCExec(
                 ['setfib', self.exec_fib, 'jexec', f'ioc-{self.uuid}']
-                + exec_start, None, None, unjailed=True, decode=True
+                + exec_start, None, uuid='', unjailed=True, decode=True
```
skarekrow: Same here

iocage_lib/ioc_stop.py
```diff
@@ -115,7 +115,7 @@ def __stop_jail__(self):
             with open(f'{self.iocroot}/log/{self.uuid}-console.log', 'a') as f:
                 with iocage_lib.ioc_exec.IOCExec(
                     ['setfib', exec_fib, 'jexec', f'ioc-{self.uuid}']
-                    + exec_stop, None, None, unjailed=True, decode=True
+                    + exec_stop, None, uuid='', unjailed=True, decode=True
```
skarekrow: and here

skarekrow (changes requested):

iocage_lib/ioc_exec.py
```diff
@@ -262,4 +262,7 @@ def exec_jail(self):
 class SilentExec(object):
     def __init__(self, *args, **kwargs):
         with IOCExec(*args, **kwargs) as silent:  # noqa
-            pass
+            self.output = list(silent)
+
+        self.stdout = ''.join([i[0] for i in self.output])
```
skarekrow, line 267: We should honor `decode` here and return byte strings by default

iocage_lib/ioc_common.py
sonicaj: Ooops

skarekrow (changes requested):

iocage_lib/ioc_exec.py
sonicaj: Actually that is being taken care of in IOCExec ?

skarekrow (approved):

Issue #783, opened 2019-01-06
Enhancement Request: option to "regenerate mac" on clone


qubitrenegade:
- [x] Supply `iocage --version`

```
root@freenas # iocage --version                                     
Version 1.0 ALPHA 1
```

- [x] Supply the commands used, along with any steps to recreate it.

```# assumes jail created named "test"
iocage create --name test --release 11.2-RELEASE vnet=on dhcp=on bpf=yes 
iocage start test
iocage stop test
iocage clone test --name clone
iocage get vnet0_mac clone
iocage set vnet0_mac="02ff60ae1b76 02ff60ae1b77" clone
iocage start clone
```

- [x] Provide the output from the command you issued.

```
root@freenas[~]# iocage create --name test --release 11.2-RELEASE vnet=on dhcp=on bpf=yes 
test successfully created!
root@freenas[~]# iocage get vnet0_mac test                                               
none
root@freenas[~]# iocage start test
* Starting test
  + Started OK
  + Configuring VNET OK
  + DHCP Address: 192.168.0.40/24
  + Starting services OK
root@freenas[~]# iocage get vnet0_mac test
02ff60ae1b75 02ff60ae1b76
root@freenas[~]# iocage stop test
* Stopping test
  + Running prestop OK
  + Stopping services OK
  + Tearing down VNET OK
  + Removing devfs_ruleset: 13 OK
  + Removing jail process OK
  + Running poststop OK
root@freenas[~]# iocage clone test --name clone
clone successfully cloned!
root@freenas[~]# iocage get vnet0_mac clone
02ff60ae1b75 02ff60ae1b76
root@freenas[~]# iocage start clone
* Starting clone
  + Started OK
  + Configuring VNET OK
  + DHCP Address: 192.168.0.40/24
  + Starting services OK
root@freenas[~]# iocage stop clone
* Stopping clone
  + Running prestop OK
  + Stopping services OK
  + Tearing down VNET OK
  + Removing devfs_ruleset: 13 OK
  + Removing jail process OK
  + Running poststop OK
root@freenas[~]# iocage set vnet0_mac="02ff60ae1b76 02ff60ae1b77"
You must specify a jail!
root@freenas[~]# iocage set vnet0_mac="02ff60ae1b76 02ff60ae1b77" clone
Property: vnet0_mac has been updated to 02ff60ae1b76 02ff60ae1b77
root@freenas[~]# iocage start clone
* Starting clone
  + Started OK
  + Configuring VNET OK
[...]

PR #784, opened 2019-01-06
Update document, include info on DHCP and MAC

qubitrenegade:
- [X] Explain the feature

Related to what I found in #783, add some language to the Networking document to explain how to setup DHCP with VNET and how to cause the MAC address to be regenerated.

Let me know if you'd like me to change anything.

Thanks!
- Q

- [X ] Read [CONTRIBUTING.md](https://github.com/iocage/iocage/blob/master/CONTRIBUTING.md)

### DIFF
--- a/doc/source/networking.rst
+++ b/doc/source/networking.rst
@@ -205,9 +205,13 @@ To configure both IPv4 and IPv6:
 
 .. note:: For VNET jails, a default route has to also be specified.
 
-To enable DHCP:
+To create a a jail with a DHCP interface add the `dhcp=on` property:
 
-:samp:`# iocage set vnet=on dhcp=on bpf=yes examplejail`
+:samp:`# iocage create -r 11.0-RELEASE --name myjail dhcp=on`
+
+The `dhcp=on` property implies creating a VNET virtual network stack and
+enabling the Berkley Packet Filter. DHCP cannot work without VNET.
+More information about VNET is available in the VNET(9) FreeBSD manual page.
 
 .. index:: Tips for configuring VNET
 .. _Tips for Configuring VNET:
--- a/doc/source/networking.rst
+++ b/doc/source/networking.rst
@@ -207,7 +207,7 @@ To configure both IPv4 and IPv6:
 
 To enable DHCP:
 
-:samp:`# iocage set vnet=on dhcp=on bpf=yes examplejail`
+:samp:`# iocage set dhcp=on examplejail`
 
 .. index:: Tips for configuring VNET
 .. _Tips for Configuring VNET:
--- a/doc/source/networking.rst
+++ b/doc/source/networking.rst
@@ -205,6 +205,10 @@ To configure both IPv4 and IPv6:
 
 .. note:: For VNET jails, a default route has to also be specified.
 
+To enable DHCP:
+
+:samp:`# iocage set vnet=on dhcp=on bpf=yes examplejail`
+
 .. index:: Tips for configuring VNET
 .. _Tips for Configuring VNET:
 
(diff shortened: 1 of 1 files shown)


doc/source/networking.rst
```diff
@@ -205,6 +205,10 @@ To configure both IPv4 and IPv6:
 
 .. note:: For VNET jails, a default route has to also be specified.
 
+To enable DHCP:
+
+:samp:`# iocage set vnet=on dhcp=on bpf=yes examplejail`
+
```
sonicaj, line 215: Actually now just setting one prop should work `dhcp=on`. This prop will automatically set the other two props to true.

qubitrenegade: You know...  I should have tried that... :)

Ok, I'll update and repush.  Thanks! :+1:

qubitrenegade: Neat!  This github browser editor is really cool.

Change updated as requested.  Do you need me to squash the commits into one?

skarekrow: Actually setting `dhcp=on` is only during creation, during `set` it won't set the others as they may already have values users care about.

qubitrenegade: ok, so just to clarify:

When creating a new clone, you can do 

```
iocage create ... dhcp=on
```

But if you're updating a container that was previously statically set, you need:

```
iocage set vnet=on dhcp=on bpf=yes
```

To ensure all of the appropriate settings are set for DHCP?

skarekrow: When creating a new jail and cloning one yes, also correct.

qubitrenegade: Ok, great.  I reverted the commit removing those from the networking doc.

I could add something like the following to the basic-use doc:

```
Creates a FreeBSD 11.0-RELEASE jail with the custom name *myjail* and
sets the jail to start at system boot time.

+ To create a a jail with a DHCP interface add the `dhcp=on` property:
+
+ :samp:`# iocage create -r 11.0-RELEASE --name myjail dhcp=on`
+
+ The `dhcp=on` implies creating a VNET virtual network stack and
+ enabling the Berkley Packet Filter. DHCP cannot work without a VNET.
+ More information about VNETs is available in the VNET(9) FreeBSD manual page.

More information about jail properties is available in the iocage(8)
```

skarekrow:
```suggestion
Creates a FreeBSD 11.0-RELEASE jail with the custom name *myjail* and
sets the jail to start at system boot time.

+ To create a a jail with a DHCP interface add the `dhcp=on` property:
+
+ :samp:`# iocage create -r 11.0-RELEASE --name myjail dhcp=on`
+
+ The `dhcp=on` property implies creating a VNET virtual network stack and
+ enabling the Berkley Packet Filter. DHCP cannot work without VNET.
+ More information about VNET is available in the VNET(9) FreeBSD manual page.

More information about jail properties is available in the iocage(8)
```

(Fixed a couple typos and added a word) Sounds good to me

qubitrenegade: Ok, I apologize for the delay here, IRL got busy and this somehow fell off my github notifications.

It's been about a month, so I'm having trouble remembering where the "Creates a FreeBSD 11.0-RELEASE jail with the custom name *myjail* and sets the jail to start at system boot time." line came from as it doesn't seem to fit in context...

Thoughts/Comments?

Thanks!

skarekrow: You know, I was just looking this over and I'm also not sure where it grabbed that change. No apologies needed, in fact in this time a new DHCP feature was added. You can actually supply `INTERFACE|dhcp` now, so for example most users will find it useful to do `vnet0|DHCP`, this only runs DHCP on *just* the vnet0 interface as opposed to running it on all interfaces (which is what dhcp=on did)

I suppose that might not be super relevant for an easy example, thoughts?

skarekrow (approved):

skarekrow:
Thanks!

PR #785, opened 2019-01-06
BUG: SYSV settings are ignored

cernst1980:
When setting sysvmsg, sysvsem and sysvshm to value of "new", jail starts without these settings.
The settings were skipped because of "if '1' in x".

- [x] Explain the feature


iocage_lib/ioc_start.py
```diff
@@ -367,6 +367,9 @@ def __start_jail__(self):
             x for x in net
             + [x for x in parameters if '1' in x]
             + [
+                _sysvmsg,
+                _sysvsem,
+                _sysvshm,
```
skarekrow, line 373: Good catch, can we follow the other formatting and move these under the name?

skarekrow (changes requested):

iocage_lib/ioc_start.py
cernst1980: Sorry, I do not get what you mean.
Can you please suggest a specific code change.

iocage_lib/ioc_start.py
```diff
@@ -367,6 +367,9 @@ def __start_jail__(self):
...
                 f'name=ioc-{self.uuid}',
```
```suggestion
                 f'name=ioc-{self.uuid}',
                 _sysvmsg,
                 _sysvsem,
                 _sysvshm,
```
Something similar to that. Not sure why Github is showing that topline change, may be the newline.

skarekrow (changes requested):

iocage_lib/ioc_start.py
skarekrow: Sure remove them from here and look at my suggestion

iocage_lib/ioc_start.py
```diff
@@ -367,7 +367,13 @@ def __start_jail__(self):
...
             + [x for x in parameters if '1' in x]
             + [
-                f'name=ioc-{self.uuid}',
+                _sysvmsg,
+                _sysvsem,
+                _sysvshm,
+                 f'name=ioc-{self.uuid}',
+                 _sysvmsg,
+                 _sysvsem,
+                 _sysvshm,
```
skarekrow: From the GH preview it looks like these have an extra whitespace in front? Make sure they line up with host.domainname below

skarekrow (changes requested):

skarekrow (approved):

mkaniewski:
Could you please merge these changes to 1.0 (stable)?

Issue #763, opened 2018-12-27
Replace 'chroot' suggestions in documentation


kpfleming:
In #542 the `chroot` command was marked as deprecated in favor of `exec` and `console`, but there are still quite a few references to it throughout the documentation in the `doc` directory. I'm happy to send a PR to replace all of them if there's any consensus on what they should be replaced with...

skarekrow:
I think it depends on the context, but I would say that most if not all situations can be replaced with `exec`.

kpfleming:
The one in `templates.rst` is going to be tricky, because it's a suggestion for modifying the content of a template, but `exec` requires starting the jail which can't be done with a template... and I don't think the `chroot` method would have worked either since the template dataset is set to read-only. I suspect the only way to fix that one is to remove it, and leave the suggestion for converting the template back to a normal jail in order to modify it.

gronke:
@kpfleming this issue with Templates was discussed with @skarekrow a couple of weeks ago. Turns out, it's one of the reasons why libiocage is no longer part of the iocage project.

This project moves jails from the `jails` to `templates` dataset when turning them into templates. As soon as a jail was created from a template, it is no longer possible to turn it back into a jail. (see https://github.com/libiocage/libiocage/issues/551#issuecomment-430959716)

Therefore libiocage marks templates as such from the jails configuration rather than moving their dataset. On the other hand it is necessary to return to GUIDs for jail storage locations, as deletion of templates requires the following procedure:

### Template deletion
When deleting a template, check if there are jails existing that have been created from them. If so (1), mark the template as deleted, otherwise delete the template.

### Jail deletion
When deleting a jail, check if it was created from a template. When the template has been marked as deleted and this jail is the only child of the template, delete it as well.

skarekrow:
> The one in `templates.rst` is going to be tricky, because it's a suggestion for modifying the content of a template, but `exec` requires starting the jail which can't be done with a template... and I don't think the `chroot` method would have worked either since the template dataset is set to read-only. I suspect the only way to fix that one is to remove it, and leave the suggestion for converting the template back to a normal jail in order to modify it.

Perhaps the best way to move forward with that snippet is to just remove it. As changing templates would unintuitively make a user expect the changes to trickle down to jails which is not the case. As it's just a parent of a snapshot.

PR #794, opened 2019-01-10
Remove references to 'chroot' command from docs

kpfleming:
The 'chroot' command has been deprecated, so the docs should not refer to it
or suggest its use. For templates, this means that the only way to modify
a template is to convert it back into a jail, change it, and then convert to
a template again.

Closes issue #763

### DIFF
--- a/doc/source/templates.rst
+++ b/doc/source/templates.rst
@@ -42,12 +42,10 @@ Done!
 
 **Customizing a template:**
 
-To make further customizations or just patch the template, there are two
-options:
+To make further customizations or just patch the template, you'll need
+to convert the template back into a jail, start it, and then convert it
+back to a template once you've completed your changes.
 
 * Convert the template back to a jail with
   :command:`iocage set template=no [UUID | NAME]`, then start the jail
   with :command:`iocage start [UUID | NAME]`.
-* If network access is unnecessary to make the changes, use
-  :command:`iocage chroot [UUID | NAME] [Command ...]` to run the
-  needed commands inside the template.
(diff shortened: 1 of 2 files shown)


doc/source/templates.rst
```diff
@@ -42,12 +42,10 @@ Done!
 
 **Customizing a template:**
 
-To make further customizations or just patch the template, there are two
-options:
+To make further customizations or just patch the template, you'll need
+to convert the template back into a jail, start it, and then convert it
+back to a template once you've completed your changes.
```
igalic: n.b.: This only works so long as no jail has been created from this template.

igalic (commented):
👀

doc/source/templates.rst
kpfleming: That's a great point. I'll rework this text, probably to indicate that if jails have already been created from the template then the best course of action would be clone the template into a new template, make the changes, and then create new jails from that.

igalic: you're describing a process that cloud providers usually call (custom) image creation, so i'd go looking through those docs if words fail you

skarekrow: I recommend using a thickjail for that clone with the new template. Otherwise it too will be dependent on the old living.

skarekrow:
@kpfleming How goes it? I think this is important to change for users who might trip on this. Thought I'd send a heartbeat so this doesn't get forgotten :)

kpfleming:
Sorry for the delay, but I've pushed an expanded set of instructions now.

skarekrow (approved):

skarekrow:
Thanks!

PR #797, opened 2019-01-11
tkt-69267: Add --debug(-d) and --images(-i) flag to clean

skarekrow:
Also switch over to subprocess zfs as occasionally the base dataset would not be destroyed properly by py-libzfs.

FreeNAS Ticket: #69267

### DIFF
--- a/iocage_lib/ioc_create.py
+++ b/iocage_lib/ioc_create.py
@@ -109,6 +109,7 @@ class IOCCreate(object):
         import iocage_lib.ioc_destroy  # Circular dep
         start = False
         is_template = False
+        source_template = None
         rtsold_enable = 'NO'
 
         if iocage_lib.ioc_common.match_to_dir(self.iocroot, jail_uuid):
@@ -151,6 +152,7 @@ class IOCCreate(object):
                     except KeyError:
                         # Thick jails won't have this
                         cloned_release = None
+                    source_template = self.release
                 elif self.clone:
                     _type = "jails"
                     clone_path = f"{self.iocroot}/{_type}/{self.release}"
@@ -211,7 +213,9 @@ class IOCCreate(object):
                 if cloned_release is None:
                     cloned_release = self.release
 
-                config = self.create_config(jail_uuid, cloned_release)
+                config = self.create_config(
+                    jail_uuid, cloned_release, source_template
+                )
             else:
                 clone_config = f"{self.iocroot}/jails/{jail_uuid}/config.json"
                 clone_fstab = f"{self.iocroot}/jails/{jail_uuid}/fstab"
@@ -609,7 +613,7 @@ class IOCCreate(object):
 
         return jail_uuid
 
-    def create_config(self, jail_uuid, release):
+    def create_config(self, jail_uuid, release, source_template):
         """
         Create the jail configuration with the minimal needed defaults.
         If self.thickconfig is True, it will create a jail with all properties.
@@ -628,6 +632,9 @@ class IOCCreate(object):
             d_conf = iocage_lib.ioc_json.IOCJson().check_default_config()
             jail_props.update(d_conf)
             jail_props['CONFIG_TYPE'] = 'THICK'
+
+        if source_template is not None:
+            jail_props['source_template'] = source_template
 
         return jail_props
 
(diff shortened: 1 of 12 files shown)


iocage_lib/ioc_create.py
```diff
@@ -628,6 +633,9 @@ def create_config(self, jail_uuid, release):
             jail_props.update(d_conf)
             jail_props['CONFIG_TYPE'] = 'THICK'
 
+        if source_template is not None:
```
sonicaj, line 636: Should we perhaps take care of this prop as well as we discussed ?

iocage_lib/ioc_destroy.py
```diff
@@ -79,147 +79,138 @@ def __stop_jails__(datasets, path=None, root=False, clean=False):
...
             except IndexError:
                 # A RELEASE dataset
                 return
 
             # We want the real path now.
-            _path = dataset.properties["mountpoint"].value.replace("/root", "")
-            # It gives us a string that says "none", not terribly
-            # useful, fixing that.
-            _path = _path if _path != "none" else None
+            _path = mountpoint.replace('/root', '')
```
sonicaj: Perhaps we think more about this as well ?

iocage_lib/ioc_destroy.py
```diff
@@ -79,147 +79,138 @@ def __stop_jails__(datasets, path=None, root=False, clean=False):
...
+                                                suppress_exception=True)
 
     def __destroy_leftovers__(self, dataset, clean=False):
         """Removes parent datasets and logs."""
-        uuid = dataset.name.rsplit("/root")[0].split("/")[-1]
-        snapshot = False
-
-        try:
-            path = dataset.properties["mountpoint"].value
+        uuid = dataset.rsplit('/root')[0].split('/')[-1]
```
sonicaj: This would be related to the above one as well
`'vol1/iocage/jails/root/root'.rsplit('/root')[0].split('/')[-1]`

iocage_lib/ioc_json.py
```diff
@@ -303,6 +322,53 @@ def zfs_get_snapshot(self, snap_id):
...
+            ).stdout.decode().split())
+        except su.CalledProcessError as e:
+            iocage_lib.ioc_common.logit(
+                {
+                    'level': 'EXCEPTION',
+                    'message': 'Getting dataset and dependents for '
+                               f'{identifier} failed!\n'
+                               f'Reason: {e.stderr.decode()}'
+                },
+                _callback=self.callback,
```
sonicaj, line 383: Would it be possible that this raises an undefined error if somewhere we use this function and callback isn't defined ?

tests/functional_tests/0013_import_test.py
```diff
@@ -61,10 +61,3 @@ def test_01_import_jail(invoke_cli, jail, skip_test, remove_file, zfs):
     )
 
     assert jail.exists is True, f'{exported_jail} jail did not import'
-
-    # Let's do a clean up on the exported files
```
sonicaj, line 65: I think we should let these be in case we are running selected tests and then import would not clean up on the images dataset ? Thoughts ?

skarekrow: I suppose we can, but we don't officially support running selected tests, right now it's the whole battery

iocage_lib/ioc_json.py
skarekrow: callback will be None in that case, which is expected most of the time

iocage_lib/ioc_create.py
skarekrow: That's a different PR to rename that, unless you're talking about something else?

sonicaj: Yup, I'm referring to renaming that. But if we don't do that, we would not be cleaning up properly, thoughts ?

iocage_lib/ioc_json.py
sonicaj: So if we use IOCZFS directly, it would not exist then

iocage_lib/ioc_create.py
sonicaj: Oh, I think I misunderstood you. Are you saying we should take care of that use case in another PR ?

skarekrow: Yep :)

sonicaj (dismissed):

sonicaj (dismissed):

sonicaj (approved):

PR #802, opened 2019-01-13
Improve exception handling

sonicaj:
This commit improves how we handle exception handling and also fixes an issue where we could not return values from existing exception handler.

### DIFF
--- a/iocage_lib/ioc_exceptions.py
+++ b/iocage_lib/ioc_exceptions.py
@@ -76,16 +76,18 @@ class ValueNotFound(Exception):
 
 
 @contextmanager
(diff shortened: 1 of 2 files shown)


skarekrow (changes requested):
SInce we are no longer using a return value, it seems to me we could drop this for contextlib's suppress and just return the values the same way, without holding onto code ourselves.

sonicaj:
I have no objections to that. Cleaning up however can still be beneficial. If you think not for now, I can drop all usages altogether then

skarekrow (dismissed):

skarekrow (dismissed):

skarekrow (approved):

PR #818, opened 2019-01-21
Improve Makefile

### LINKED ISSUE #813: Python3 missing when installing on clean system
When installing iocage from the sources on a clean system, `make install` fails because Python 3 is missing.

### Current Result
```
root@iocage:/usr/local/src/iocage # make install
Python binary /usr/local/bin/python3 not found, please enter a valid one
*** Error code 1

Stop.
make: stopped in /usr/local/src/iocage
```

### Expected Result
The `python36` package is installed if it is missing (among with `py36-libzfs`).

### LINKED ISSUE #814: Installation requires latest pkg sources
When attempting to install iocage using `make install` (as recommended in the readme), the installer requires the `latest` package source to be used by the host. This breaks with custom package sources, such as poudriere. Also this changes have a global effect on the host and other installed packages and therefore should not be mandatory.

### Current Result
```console
root@iocage:/usr/local/src/iocage # make install
Please ensure pkg url is using "latest" instead of "quarterly" in /etc/pkg/FreeBSD.conf before proceeding with installation
*** Error code 1

Stop.
make: stopped in /usr/local/src/iocage
```

### Expected Result
The installation continues regardless of the configured or enabled pkg sources.

### LINKED ISSUE #815: Installation installs Python 2 dependencies
When installing iocage using `make install`, Python 2 packages are installed among the required Python 3 dependencies of iocage.

### Current Result
The command is taken from https://github.com/iocage/iocage/blob/679b5674f6d8969c3ef0d0277607af573d6748eb/Makefile#L9

```console
root@iocage:/usr/local/src/iocage # pkg install -y devel/py-libzfs                                                                                                                                                                                                            
Updating FreeBSD repository catalogue...
FreeBSD repository is up to date.
All repositories are up to date.
The following 6 package(s) will be affected (of 0 checked):

New packages to be INSTALLED:
        py27-libzfs: 1.0.20181220 [FreeBSD]
        py36-libzfs: 1.0.20181220 [FreeBSD]
        python27: 2.7.15 [FreeBSD]
        py27-setuptools: 40.6.2 [FreeBSD]
        py27-enum34: 1.1.6 [FreeBSD]
        py36-setuptools: 40.6.2 [FreeBSD]

Number of packages to be installed: 6
```

### Expected Result
Because iocage requires Python3 exclusively, no Python 2 dependencies should be installed.

sonicaj:
This commit introduces the following changes:
1) Installs python36 if python provided path does not exist
2) Removes a compulsion for pkg to be using latest instead of quarterly with a warning
3) Make sure we don't install any python2 deps

FreeNAS Ticket: #71103
Closes #813
Closes #814
Closes #815

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,16 @@
 ZPOOL=""
 SERVER=""
-PYTHON?=/usr/local/bin/python3
+PYTHON?=/usr/local/bin/python3.6
 
-install:
-	@test -s ${PYTHON} || (echo "Python binary ${PYTHON} not found, please enter a valid one"; exit 1)
-	@(pkg -vv | grep -e "url.*/latest") > /dev/null 2>&1 || (echo "Please ensure pkg url is using \"latest\" instead of \"quarterly\" in /etc/pkg/FreeBSD.conf before proceeding with installation"; exit 1)
+depends:
+	@(pkg -vv | grep -e "url.*/latest") > /dev/null 2>&1 || (echo "It is advised pkg url is using \"latest\" instead of \"quarterly\" in /etc/pkg/FreeBSD.conf.";)
+	@test -s ${PYTHON} || (echo "Python binary ${PYTHON} not found, iocage will install python36"; pkg install -q -y python36)
+	pkg install -q -y py36-libzfs
 	${PYTHON} -m ensurepip
-	pkg install -y devel/py-libzfs
-	${PYTHON} -m pip install -Ur requirements.txt .
+	${PYTHON} -m pip install -Ur requirements.txt
+
+install: depends
+	${PYTHON} -m pip install -U .
 uninstall:
 	${PYTHON} -m pip uninstall -y iocage-lib iocage-cli
 test:


Makefile
```diff
@@ -1,13 +1,16 @@
 ZPOOL=""
 SERVER=""
-PYTHON?=/usr/local/bin/python3
+PYTHON?=/usr/local/bin/python3.6
 
-install:
-	@test -s ${PYTHON} || (echo "Python binary ${PYTHON} not found, please enter a valid one"; exit 1)
-	@(pkg -vv | grep -e "url.*/latest") > /dev/null 2>&1 || (echo "Please ensure pkg url is using \"latest\" instead of \"quarterly\" in /etc/pkg/FreeBSD.conf before proceeding with installation"; exit 1)
+depends:
+	@(pkg -vv | grep -e "url.*/latest") > /dev/null 2>&1 || (echo "It is advised pkg url is using \"latest\" instead of \"quarterly\" in /etc/pkg/FreeBSD.conf.";)
```
skarekrow, line 6: This was originally added since python3.6 was only in latest, I believe this is in quarterly now, so we can probably just drop this test

skarekrow (changes requested):

Makefile
sonicaj: Actually there's another usecase - py-libzfs. Some tests would fail if py-libzfs is being used with the quarterly version. There still might be some bugs as well which have been addressed in latest ( an assumption for the bug one )

sonicaj: Please advise on how to proceed. I think we should keep this warning in this case

skarekrow (approved):

Makefile
skarekrow: Agreed :)

PR #822, opened 2019-01-22
tkt-71587: Fix erroneous template messages on creation failure

### LINKED ISSUE #809: Can't re-use jail names when creating with a template

# iocage --version
Version	1.0 RELEASE 2018/12

 a stable release.

# iocage create -t base -n foo
foo successfully created!

# iocage destroy foo

This will destroy jail foo

Are you sure? [y/N]: y
Destroying foo

# iocage create -t base -n foo
Template: foo not found!

skarekrow:
FreeNAS Ticket: #71587
Closes #809

### DIFF
--- a/iocage_cli/create.py
+++ b/iocage_cli/create.py
@@ -30,6 +30,7 @@ import uuid
 import click
 import iocage_lib.ioc_common as ioc_common
 import iocage_lib.iocage as ioc
+import iocage_lib.ioc_exceptions as ioc_exceptions
 
 __rootcmd__ = True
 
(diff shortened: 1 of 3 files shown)


iocage_cli/create.py
```diff
@@ -179,20 +180,23 @@ def cli(release, template, count, props, pkglist, basejail, thickjail, empty,
                       template=template, short=short, _uuid=_uuid,
                       basejail=basejail, thickjail=thickjail, empty=empty,
                       thickconfig=thickconfig)
-    except RuntimeError as err:
-        if template and "Dataset" in str(err):
+    except (RuntimeError, ioc_exceptions.JailMissingConfiguration) as err:
+        if template and "Dataset" in str(err) or str(
+                err).startswith('Template'):
```
sonicaj, line 185: 8 spaces

sonicaj: Should we perhaps do a case insensitive comparison ?

sonicaj (approved):
LGTM, just a couple of notes whether we should look at those right now

iocage_cli/create.py
skarekrow: 8 spaces are for flake, since we're line breaking with `str(` and no, the message will be capitalized.

skarekrow:
Unrelated test failure

PR #825, opened 2019-01-23
Migrate ip4 and ip6 from none to disable

skarekrow:
- Add new method that will fix other properties we may need to address in the future

FreeNAS Ticket: #69276


sonicaj (approved):

skarekrow:
Unrelated test failure

PR #826, opened 2019-01-24
tkt-72067: Uppercase the RELEASE specified during create

skarekrow:
This helps avoid lazy sysadmin syndrome

Reported by: Michael W Lucas

FreeNAS Ticket: #72067

### DIFF
--- a/iocage_lib/iocage.py
+++ b/iocage_lib/iocage.py
@@ -551,8 +551,8 @@ class IOCage(ioc_json.IOCZFS):
                 silent=self.silent)
 
         if not os.path.isdir(
-                f"{self.iocroot}/releases/{release}") and not template and \
-                not empty and not clone:
+            f'{self.iocroot}/releases/{release}'
+        ) and not template and not empty and not clone:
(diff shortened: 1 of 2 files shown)


iocage_lib/iocage.py
```diff
@@ -550,9 +550,14 @@ def create(self,
                 _callback=self.callback,
                 silent=self.silent)
 
+        if os.path.isdir(
+            f'{self.iocroot}/releases/{release.upper()}'
```
sonicaj: Should we perhaps to be on the safe side do something like `(release or '').upper()` for someone who might be using lib directly ?

sonicaj (dismissed):
LGTM, just a minor note

iocage_lib/iocage.py
skarekrow: That would trigger a false positive as `f'{self.iocroot}/releases/` almost always exists

sonicaj: that is true, i missed that - maybe in this case we should just check release is true

sonicaj (approved):

sonicaj:
Thank you ;)

PR #833, opened 2019-01-28
tkt-70993: Fix hostid_strict_check property value

skarekrow:
FreeNAS Ticket: #70993


skarekrow:
Test failure is a false positive

sonicaj (approved):

PR #837, opened 2019-01-29
tkt-72731: Allow regeneration of a new mac during clone (--newmac|-N)

### LINKED ISSUE #783: Enhancement Request: option to "regenerate mac" on clone

skarekrow:
- Cleanup snapshots after clone failure
- Correct a capital character

FreeNAS Ticket: #72731
Closes #783

### DIFF
--- a/iocage_cli/clone.py
+++ b/iocage_cli/clone.py
@@ -26,7 +26,6 @@ import click
 
 import iocage_lib.ioc_common as ioc_common
 import iocage_lib.iocage as ioc
-import subprocess as su
 
 __rootcmd__ = True
 
--- a/iocage_cli/clone.py
+++ b/iocage_cli/clone.py
@@ -69,14 +69,6 @@ def cli(source, props, count, name, _uuid, thickjail, newmac):
         for p in ('vnet0', 'vnet1', 'vnet2', 'vnet3'):
             props.append(f'{p}_mac=none')
 
-    try:
-        ioc.IOCage(jail=source, skip_jails=True).create(
-            source, props, count, _uuid=_uuid, thickjail=thickjail, clone=True
-        )
-    except BaseException:
-        # SystemExit or other friends, we don't care, just want to cleanup
-        pool = ioc.PoolAndDataset().get_pool()
-        su.run(
-            ['zfs', 'destroy', '-r', f'{pool}/iocage/jails/{source}@{_uuid}']
-         )
-        raise
+    ioc.IOCage(jail=source, skip_jails=True).create(
+        source, props, count, _uuid=_uuid, thickjail=thickjail, clone=True
+    )
--- a/iocage_cli/clone.py
+++ b/iocage_cli/clone.py
@@ -26,6 +26,7 @@ import click
 
 import iocage_lib.ioc_common as ioc_common
 import iocage_lib.iocage as ioc
+import subprocess as su
 
 __rootcmd__ = True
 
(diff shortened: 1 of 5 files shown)


iocage_cli/clone.py
```diff
@@ -52,21 +53,30 @@ def validate_count(ctx, param, value):
...
-        ioc_common.logit({
-            'level': 'EXCEPTION',
-            'message': msg
-        })
+    try:
+        ioc.IOCage(jail=source, skip_jails=True).create(
+            source, props, count, _uuid=_uuid, thickjail=thickjail, clone=True
+        )
+    except BaseException:
+        # SystemExit or other friends, we don't care, just want to cleanup
```
sonicaj: I think we should do clean up in iocage lib as that will ensure that anyone using the lib won't have leftovers. Thoughts ?

iocage_lib/iocage.py
```diff
@@ -645,7 +645,14 @@ def create(self,
                     clone=clone,
                     thickconfig=thickconfig
                 ).create_jail()
-        except RuntimeError:
+        except BaseException:
+            if clone:
+                su.run(
```
themylogin, line 650: Is this going to be used on CLI and you catch `BaseException` because you want to catch Ctrl-C? If so, some logging can be useful, explaining why there is no immediate reaction to it.

themylogin (approved):

iocage_lib/iocage.py
skarekrow: `BaseException` is used to catch `SystemExit`, and all the other `Exception` inherited exceptions iocage can throw. `KeyboardInterrupt` is actually handled in `ioc_create.py`

PR #842, opened 2019-02-01
tkt-73236: Make booleans universal values

skarekrow:
This greatly simplifies the user experience, in addition allows more boolean type properties to be easier added in the future with less regard to their parsing.

- All boolean properties can now be set as '1/0', 'on/off', 'yes/no', 'true/false'
- These now rest on disk as 1/0
- Fix some IP validation issues

FreeNAS Ticket: #73236

### DIFF
--- a/iocage_lib/ioc_create.py
+++ b/iocage_lib/ioc_create.py
@@ -310,11 +310,15 @@ class IOCCreate(object):
             # Clones are expected to be as identical as possible.
 
             for k, v in config.items():
-                v = v.replace(clone_uuid, jail_uuid)
+                try:
+                    v = v.replace(clone_uuid, jail_uuid)
 
-                if '_mac' in k:
-                    # They want a unique mac on start
-                    config[k] = 'none'
+                    if '_mac' in k:
+                        # They want a unique mac on start
+                        config[k] = 'none'
+                except AttributeError:
+                    # Bool props
+                    pass
 
                 config[k] = v
         else:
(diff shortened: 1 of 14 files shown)


iocage_lib/ioc_create.py
```diff
@@ -396,49 +401,56 @@ def _create_jail(self, jail_uuid, location):
...
+                            ' enabling!'
                         },
                             _callback=self.callback,
                             silent=self.silent)
-                        config['vnet'] = 'on'
+                        config['vnet'] = 1
 
                     rtsold_enable = 'YES'
-            elif (key == 'dhcp' and value == 'on') or (
+            elif (key == 'dhcp' and is_true) or (
```
sonicaj, line 417: So I think we left a case here. Suppose we have an `ip4addr` with the value `vnet0|DHCP,vtnet0|192.168.10.10`. In this case we would like DHCP to be enabled on vnet0 but also that we have the ip `192.168.10.10` configured on `vtnet0`. If both are tried separately they will work but this fails collectively. We should check this differently, thoughts ?

iocage_lib/ioc_json.py
```diff
@@ -1939,11 +1987,18 @@ def json_check_prop(self, key, value, conf):
                     return value, conf
 
             if props[key][0] == 'string':
-                if key in ('ip4_addr', 'ip6_addr') and value != 'none':
+                if key in (
+                    'ip4_addr', 'ip6_addr'
+                ) and (
+                    value != 'none' and 'DHCP' not in value.upper() and
```
sonicaj, line 1993: Same concern as above for the ip values

iocage_lib/ioc_list.py
```diff
@@ -126,11 +126,13 @@ def list_all_quick(self, jails):
 
             uuid = conf["host_hostuuid"]
             ip4 = conf.get('ip4_addr', 'none')
-            dhcp = True if conf.get('dhcp') == 'on' or 'DHCP' in \
-                ip4.upper() else False
+            dhcp = True if iocage_lib.ioc_common.check_truthy(
```
sonicaj, line 129: This is tricky. I am not sure what is the best way forward with this. This also follows the same pattern as above for ip values.

sonicaj (commented):
Good work ;) Just a few thoughts

iocage_lib/ioc_create.py
skarekrow: That case cannot happen, since it's VNET, you can only set it on the vnet0 adapter

sonicaj: As discussed we can take care of this in another PR as this is out of scope for this one ;)

sonicaj (approved):
Thank you :)

Issue #841, opened 2019-01-31
"iocage create" crash


agshekeloh:
<pre>

Today's iocage, yesterday's -current and packages

storm~;iocage --version
Version 1.1 RELEASE 2019/01
storm~;uname -a
FreeBSD storm 13.0-CURRENT FreeBSD 13.0-CURRENT r343578 GENERIC  amd64

storm~;iocage create -n squid1 -r 12.0-RELEASE
Traceback (most recent call last):
  File "/usr/local/bin/iocage", line 10, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/iocage_cli/create.py", line 182, in cli
    thickconfig=thickconfig)
  File "/usr/local/lib/python3.6/site-packages/iocage_lib/iocage.py", line 646, in create
    thickconfig=thickconfig
  File "/usr/local/lib/python3.6/site-packages/iocage_lib/ioc_create.py", line 92, in create_jail
    return self._create_jail(jail_uuid, location)
  File "/usr/local/lib/python3.6/site-packages/iocage_lib/ioc_create.py", line 175, in _create_jail
    self.release
  File "/usr/local/lib/python3.6/site-packages/iocage_lib/ioc_common.py", line 899, in get_jail_freebsd_version
    for line in r:
  File "/usr/local/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 47: ordinal not in range(128)

Expected output? "Here's your jail."

skarekrow:
What does your env and locale look like?

agshekeloh:
storm~;env
HOME=/home/mwlucas
[...]

PR #845, opened 2019-02-04
Correct Locale settings

sonicaj:
This commit forces locale settings to be 'en_US.UTF-8' for iocage fixing any unintended issues with user's locale settings.


sonicaj:
Closes: #841

skarekrow (approved):

PR #847, opened 2019-02-05
tkt-73740: Add ip_hostname support to iocage

skarekrow:
If a user sets their DNS records up and they match the jail name, this bool will allow the user to skip configuring the IP manually, as jail(8) will grab the IP from the resolver.

FreeNAS Ticket: #73740

### DIFF
--- a/iocage_lib/ioc_create.py
+++ b/iocage_lib/ioc_create.py
@@ -611,11 +611,15 @@ class IOCCreate(object):
                 silent=self.silent)
 
         if self.pkglist:
-            if config.get('ip4_addr', 'none') == 'none' and \
-                config.get('ip6_addr', 'none') == 'none' and \
-                    not iocage_lib.ioc_common.check_truthy(
-                        config.get('dhcp', 0)
-            ):
+            dhcp_or_hostname = iocage_lib.ioc_common.check_truthy(
+                config.get('dhcp', 0)
+            ) or iocage_lib.ioc_common.check_truthy(
+                config.get('ip_hostname', 0)
+            )
+
+            if config.get('ip4_addr', 'none') == "none" and \
+                config.get('ip6_addr', 'none') == "none" and \
+                    not dhcp_or_hostname:
                 iocage_lib.ioc_common.logit({
                     "level": "WARNING",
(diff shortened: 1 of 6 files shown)


sonicaj (dismissed):

themylogin (dismissed):

iocage_lib/ioc_create.py
```diff
@@ -611,11 +611,15 @@ def _create_jail(self, jail_uuid, location):
                 silent=self.silent)
 
         if self.pkglist:
-            if config.get('ip4_addr', 'none') == 'none' and \
-                config.get('ip6_addr', 'none') == 'none' and \
-                    not iocage_lib.ioc_common.check_truthy(
-                        config.get('dhcp', 0)
-            ):
+            dhcp_or_hostname = iocage_lib.ioc_common.check_truthy(
+                config.get('dhcp', 0)
```
sonicaj: I might be wrong here, but umm isn't the config supposed to have `0` or `1` ?

sonicaj (dismissed):
Looks good, just a minor note

iocage_lib/ioc_create.py
skarekrow: It *may* not exist as a key at this point, so using get for safety.

sonicaj: But if it does, it will be `0` or `1` ?

skarekrow: Right, so if the key doesn't exist we set 0 as a safety net.

sonicaj: Perfect. So if it exists, i don't think we need use check_truthy in that case ? We can safely perhaps do `if config.get(key, default)` ?

skarekrow: The user may be supplying this property and it could be a variation of the bools, that's where `check_truthy` comes in

sonicaj: Ah I see. I thought we maybe had parsed config at this point and it contained `0` or `1` only for boolean based keys. So the config is normalized after this then ?

skarekrow:
network focused, tests aren't geared for that in cirrus

PR #848, opened 2019-02-06
tkt-73560: RCTL for Iocage

sonicaj:
This commit introduces following changes:
1) A class to manage RCTL related operations which are:
	a) Validate RCTL props
	b) Set RCTL rules
	c) Remove RCTL rules
	d) Check for existing RCTL rules
2) Ability to set RCTL props in the format 'action=amount'.
3) Setting RCTL rules after other start related operations are complete for a jail
4) Removing RCTL rules for a jail before we start any stopping operation for a jail when a jail stop is desired
5) The motivation behind 3 and 4 is that we should make sure start/stop behaviour is not affected by any RCTL rules at least for now as it can possibly affect the start/stop operations.
6) Allow setting/removing of RCTL rules for a jail which is running
7) There could be potentially more actions related to a single resource in RCTL rules, for now we only allow one. We can introduce support for multiple actions later gauging user reaction/usage.

Ticket: #73560

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -162,6 +162,152 @@ class JailRuntimeConfiguration(object):
             )
 
 
+class IOCRCTL(object):
+
+    types = {
+        'cputime', 'datasize', 'stacksize', 'coredumpsize',
+        'memoryuse', 'memorylocked', 'maxproc', 'openfiles',
+        'vmemoryuse', 'pseudoterminals', 'swapuse', 'nthr',
+        'msgqqueued', 'msgqsize', 'nmsgq', 'nsem', 'nsemop',
+        'nshm', 'shmsize', 'wallclock', 'pcpu', 'readbps',
+        'writebps', 'readiops',  'writeiops'
+    }
+
+    def __init__(self, name):
+        self.jail_name = f'ioc-{name}'
+
+    def set_rctl_rules(self, props):
+        # We expect props to be a list of tuples or a tuple
+        if not isinstance(props, list):
+            props = [props]
+
+        failed = set()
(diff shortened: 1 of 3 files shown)


iocage_lib/ioc_json.py
```diff
@@ -162,6 +162,152 @@ def __write_file(self):
...
+                            'properties '
+                            '"readbps, writebps, readiops, writeiops"'
+                        }
+                    )
+
+                if prop == 'pcpu' and int(value.split('=')[1]) > 100:
+                    iocage_lib.ioc_common.logit(
+                        {
+                            'level': 'EXCEPTION',
+                            'message': 'Pcpu property requires a valid '
```
skarekrow:
```suggestion
                            'message': 'pcpu property requires a valid '
```

iocage_lib/ioc_stop.py
```diff
@@ -87,6 +87,23 @@ def __stop_jail__(self):
...
+        if rctl_jail.rctl_rules_exist():
+            failed = rctl_jail.remove_rctl_rules()
+            if failed:
+                msg = 'Failed to remove'
+            else:
+                msg = 'Successfully removed'
+
+            iocage_lib.ioc_common.logit(
+                {
+                    'level': 'INFO',
```
skarekrow: This should become an error if they failed to be removed

skarekrow (changes requested):

iocage_lib/ioc_stop.py
sonicaj: I think I can have this covered by the force flag, so if for any reason they fail, force would do an override anyways

skarekrow (approved):

PR #849, opened 2019-02-06
tkt-74020: Remove unprotected devfs_rulesets

skarekrow:
This conditional was incorrect after recent changes to grabbing the dynamic ruleset

FreeNAS Ticket: #74020


themylogin (dismissed):

sonicaj (commented):
I think there is potential for error as int(devfs) can fail as strings are allowed as well here

skarekrow:
> I think there is potential for error as int(devfs) can fail as strings are allowed as well here

How so? They're always ints

sonicaj:
`
root@freebsd12vm:~ # iocage set devfs_ruleset=asdf foo
devfs_ruleset: 4 -> asdf
root@freebsd12vm:~ #
`

sonicaj:
Oops, sorry. Messed up the indentation. The root cause is different here though as i think we should validate that prop ? Thoughts ?

skarekrow:
Different problem, this will be grabbing the ruleset from `jls`, but that is absolutely a good candidate for validation.

sonicaj (dismissed):

sonicaj (approved):

PR #851, opened 2019-02-07
tkt-74281: Use hosts pkgs to remove pkgs for plugins

skarekrow:
- Make revoked directory on plugin install
- Remove unused variable

FreeNAS Ticket: #74281

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -191,7 +191,7 @@ class IOCRCTL(object):
                     ],
                     None, unjailed=True, decode=True
                 )
-            except ioc_exceptions.CommandFailed as e:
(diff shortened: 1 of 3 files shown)


sonicaj (commented):
I think we have a leftover usage of update which needs to be updated as we don't pass jid there at upgrade ?

skarekrow:
Where?

sonicaj:
Line 1092 ioc_plugin

skarekrow:
That's unrelated

sonicaj:
I think it will fail as we don't pass jid there and it's not a default argument in update function so that will raise an exception ?

sonicaj (approved):

skarekrow:
Unrelated failure, merging.

PR #852, opened 2019-02-08
tkt-74235: Add Cpuset support to Iocage

sonicaj:
This commit introduces the following changes:
1) A class to manage Cpuset prop wrt
  a) Validation
  b) Setting Cpuset Prop
2) Ability to set cpuset prop conforming to cpuset(1) format for '-l' flag only.
3) We only allow a very basic version of cpuset(1) support in which a user can specify a list of CPUs to apply to the jail.
4) When a jail is started, we set the cpuset prop in the system only at that point allowing the user to change the prop when the jail is running but not reflecting that change in the system. That change is only reflected in the system when jail is restarted. Motivation behind this step is that we are more likely to fail setting the prop when the jail is running in the system because of possible deadlock/validation issues raised by cpuset(1).
5) Point 3 and 4 can be revisited later to allow more support or enhance the functionality gauging user reaction/usage.

Ticket: #74235

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -162,6 +162,85 @@ class JailRuntimeConfiguration(object):
             )
 
 
+class IOCCpuset(object):
+
+    def __init__(self, name):
+        self.jail_name = f'ioc-{name}'
+
+    def set_cpuset(self, value=None):
+        if not value:
+            value = 'all'
+
+        failed = False
+        try:
+            iocage_lib.ioc_exec.SilentExec(
+                ['cpuset', '-l', value, '-j', self.jail_name],
+                None, unjailed=True, decode=True
+            )
+        except iocage_lib.ioc_exceptions.CommandFailed:
+            failed = True
+        finally:
+            return failed
+
+    @staticmethod
+    def retrieve_cpu_sets():
+        cpu_sets = -2
+        try:
+            output = iocage_lib.ioc_exec.SilentExec(
+                ['cpuset', '-g', '-s', '0'],
+                None, unjailed=True, decode=True
+            )
+        except iocage_lib.ioc_exceptions.CommandFailed:
+            pass
+        else:
+            result = re.findall(
+                r'.*mask:.*(\d+)$',
+                output.stdout.split('\n')[0]
+            )
+            if result:
+                cpu_sets = int(result[0])
+        finally:
+            return cpu_sets
+
+    @staticmethod
+    def validate_cpuset_prop(value, raise_error=True):
+        failed = False
+        cpu_sets = IOCCpuset.retrieve_cpu_sets() + 1
+
+        if not any(
+            cond for cond in (
+                re.findall(
+                    fr'^(?!.*(\b\d+\b).*\b\1\b)'
+                    fr'((?:{"|".join(map(str, range(cpu_sets)))})'
+                    fr'(,(?:{"|".join(map(str, range(cpu_sets)))}))*)?$',
+                    value
+                ),
+                value in ('off', 'all'),
+                re.findall(
+                    fr'^(?:{"|".join(map(str, range(cpu_sets - 1)))})-'
+                    fr'(?:{"|".join(map(str, range(cpu_sets)))})$',
+                    value
+                ) and int(value.split('-')[0]) < int(value.split('-')[1])
+            )
+        ):
+            failed = True
+
+        if failed and raise_error:
+            iocage_lib.ioc_common.logit(
+                {
+                    'level': 'EXCEPTION',
+                    'message': 'Please specify a valid format for cpuset '
+                               'value.\nFollowing 4 formats are supported:\n'
+                               '1) comma delimited string i.e 0,1,2,3\n'
+                               '2) a range of values i.e 0-2\n'
+                               '3) "all" - all would mean using all cores\n'
+                               '4) off'
+                }
+            )
+        else:
+            return failed
+
+
 class IOCRCTL(object):
 
     types = {
@@ -1931,6 +2010,17 @@ class IOCJson(IOCConfiguration):
                 else:
                     key = key.replace("_", ".")
 
+                if key == 'cpuset':
+                    iocage_lib.ioc_common.logit(
+                        {
+                            'level': 'INFO',
+                            'message': 'cpuset changes '
+                                       'require a jail restart'
+                        },
+                        _callback=self.callback,
+                        silent=self.silent
+                    )
+
                 # Let's set a rctl rule for the prop if applicable
                 if key in IOCRCTL.types:
                     rctl_jail = IOCRCTL(conf['host_hostuuid'])
@@ -2096,7 +2186,7 @@ class IOCJson(IOCConfiguration):
             "allow_vmm": truth_variations,
             "vnet_interfaces": ("string", ),
             # RCTL limits
-            "cpuset": ("off", "on"),
+            "cpuset": ('string',),
             "rlimits": ("off", "on"),
             "memoryuse": ('string',),
             "memorylocked": ('string',),
@@ -2309,6 +2399,8 @@ class IOCJson(IOCConfiguration):
                         )
                 elif key in IOCRCTL.types:
                     IOCRCTL.validate_rctl_props(key, value)
+                elif key == 'cpuset':
+                    IOCCpuset.validate_cpuset_prop(value)
 
                 return value, conf
             else:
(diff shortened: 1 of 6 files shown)


iocage_lib/ioc_start.py
```diff
@@ -656,6 +656,12 @@ def __start_jail__(self):
                 _callback=self.callback,
                 silent=self.silent)
 
+        self.set(
```
skarekrow, line 659: We should make sure to set this very last. Since an exception could happen.

iocage_lib/ioc_start.py
```diff
@@ -688,9 +694,23 @@ def __start_jail__(self):
...
+                'message': f'  + Setting cpuset prop'
+            })
+
+            cpuset_jail = iocage_lib.ioc_json.IOCCpuset(self.uuid)
+            cpuset_jail.validate_cpuset_prop(cpuset)
+
+            failed = cpuset_jail.set_cpuset(cpuset)
+            if failed:
+                iocage_lib.ioc_common.logit({
+                    'level': 'INFO',
```
skarekrow: ERROR?

skarekrow (changes requested):
I think point #4 we should not allow. Like other props that can't be changed while running (vnet), we should tell the user as they may naively expect it all to change the moment they set it like other props

skarekrow:
Added @themylogin to also peek

iocage_lib/ioc_start.py
sonicaj: I think we should let it remain like this because even if in case an exception is raised wrt rctl or cpuset, we don't stop the jail

iocage_lib/ioc_json.py
```diff
@@ -162,6 +162,85 @@ def __write_file(self):
...
+            return cpu_sets
+
+    @staticmethod
+    def validate_cpuset_prop(value, raise_error=True):
+        failed = False
+        cpu_sets = IOCCpuset.retrieve_cpu_sets() + 1
+
+        if not any(
+            cond for cond in (
+                re.findall(
```
themylogin, line 212: I fail to understand what's going on here, would you mind adding a few unit tests to this method to demonstrate and verify all the cases that can happen?

iocage_lib/ioc_start.py
skarekrow: Fair enough :)

iocage_lib/ioc_json.py
sonicaj: @themylogin I am validating the prop wrt the 4 rules mentioned at line 232 in my changes. And no worries, I can add a few unit tests to demonstrate the scenarios. In words for the first regex, they are as follows:
1) Make sure user doesn't duplicate cpu numbers
2) The order wrt comma's is always correct i.e `0,2` and not `0,2,`
3) Also cpu numbers supplied are valid wrt the system executing this

sonicaj: @themylogin I have added a few scenarios for functional tests as those can be tested in CI as well. Please have a look and let me know if any changes are required. Thank you

iocage_lib/ioc_json.py
```diff
@@ -1931,6 +2010,17 @@ def json_set_value(self, prop, _import=False, default=False):
                 else:
                     key = key.replace("_", ".")
 
+                if key == 'cpuset':
+                    iocage_lib.ioc_common.logit(
+                        {
+                            'level': 'INFO',
+                            'message': 'cpuset changes to be implemented '
```
skarekrow: Let's drop the `to be implemented` part of this string. So it just reads `cpuset changes require a jail restart`

iocage_lib/ioc_start.py
```diff
@@ -683,14 +689,28 @@ def __start_jail__(self):
...
 
-        self.set(
-            "last_started={}".format(datetime.datetime.utcnow().strftime(
-                "%F %T")))
+        cpuset = self.conf.get('cpuset', 'off')
+        if cpuset != 'off':
+            # Let's set the specified rules
+            iocage_lib.ioc_common.logit({
+                'level': 'INFO',
+                'message': f'  + Setting cpuset prop'
```
```suggestion
                'message': f'  + Setting cpuset to: {cpuset}'
```

iocage_lib/ioc_start.py
```diff
@@ -683,14 +689,28 @@ def __start_jail__(self):
...
+            })
+
+            cpuset_jail = iocage_lib.ioc_json.IOCCpuset(self.uuid)
+            cpuset_jail.validate_cpuset_prop(cpuset)
+
+            failed = cpuset_jail.set_cpuset(cpuset)
+            if failed:
+                iocage_lib.ioc_common.logit({
+                    'level': 'ERROR',
+                    'message': f'  + Failed to set cpuset prop'
```
```suggestion
                    'message': f'  + Failed to set cpuset to: {cpuset}'
```

tests/conftest.py
```diff
@@ -322,3 +322,15 @@ def _default_jails(resources, **kwargs):
         ]
 
     return _default_jails
+
+
+@pytest.fixture
+def run_console():
+    def _run_console(cmd, shell=False):
```
skarekrow: Let's drop the shell var. We should strive to never use it anyways.

skarekrow (changes requested):

iocage_lib/ioc_json.py
themylogin: @sonicaj I suggested unit tests specifically targeted the cpuset value parser, because it's complicated write-only regex. I don't mind them if they work, it's just I want to be safe if I'll need to change them in the future.

sonicaj: @themylogin no worries, i'll have the PR updated wrt that.

sonicaj: @themylogin I've updated the PR, please have a look and let me know if i could perhaps improve the tests ? Thank you

skarekrow (dismissed):

tests/unit_tests/1001_lib_cpuset_validation_test.py
```diff
@@ -0,0 +1,117 @@
...
+# 1) 0,1,2,3
+# 2) 0-3
+# 3) all
+# 4) off
+
+cpu_sets = IOCCpuset.retrieve_cpu_sets() + 1
+
+# Regex for point 1
+
+list_regex = re.compile(
```
themylogin: Tests should not have any condition logic, loops and, of course, regexes. Test should be as simple as possible, simple statements, e.g. `assert 1 + 1 == 2`.

tests/unit_tests/1001_lib_cpuset_validation_test.py
```diff
@@ -0,0 +1,117 @@
...
+
+from iocage_lib.ioc_json import IOCCpuset
+
+# For cpuset props we would like to test the following scenarios
+# 1) 0,1,2,3
+# 2) 0-3
+# 3) all
+# 4) off
+
+cpu_sets = IOCCpuset.retrieve_cpu_sets() + 1
```
themylogin: Unit test should never depend on execution environment. Also, you are restricting yourself to the number of CPUs test executor has. What if you need to test behavior with 1 CPU? With 128 CPUs? As we are testing only logic in `validate_cpuset_prop`, it is absolutely no important how many CPUs do we really have, only what `retrieve_cpu_sets` returns. You have to use `unittest.module` to fake it's return value in each test.

tests/unit_tests/1001_lib_cpuset_validation_test.py
```diff
@@ -0,0 +1,117 @@
...
+
+    value = f'0-{cpu_sets + 2}'
+    assert len(re.findall(range_regex, value)) == 0
+
+    assert IOCCpuset.validate_cpuset_prop(value, False) is True
+
+
+# Tests for point 3 and 4
+def test_09_off_value():
+    assert IOCCpuset.validate_cpuset_prop('off', False) is False
```
themylogin, line 105: This test is a good test!

themylogin (changes requested):
You've almost got the idea of what unit tests are but there still is some work to do. I've written a couple of suggestions above and a test template to help you get started:

```
from unittest.mock import Mock, patch
with patch("iocage.ioc_json.IOCCpuset.retrieve_cpu_sets", Mock(return_value=4)):
    assert IOCCpuset("").validate_cpuset_prop('0-1', False) is False

```

P.S. If you want a complete and structured information on this topic I personally recommend "Roy Osherove - The Art of Unit Testing"

sonicaj:
@themylogin I've updated the tests. Please have a look. Thank you for walking me through this and I have downloaded the book as well, will be digging deeper into it :)

themylogin (approved):
LGTM, thank you!

PR #862, opened 2019-02-14
tkt-72323: Drop dulwich dependency, add docurl support to ui.json for plugins

skarekrow:
- Replace Dulwich with GitPython
We were experiencing bad attribute usage with Dulwich, and while upstream has a commit to fix this, it didn't solve for our situation. Since we've had a couple problems with Dulwich in the past, let's try switching libraries to something that has been around longer.

- Also added docurl support to ui.json, this allows a plugin author to additionally specify where a user might be able to view documentation from the program authors themselves.

FreeNAS Ticket: #72323

### DIFF
--- a/iocage_lib/ioc_plugin.py
+++ b/iocage_lib/ioc_plugin.py
@@ -1338,7 +1338,7 @@ fingerprint: {fingerprint}
                 },
                 _callback=self.callback)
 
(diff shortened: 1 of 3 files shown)


skarekrow:
Marking WIP, still has a small bit left

sonicaj (approved):

PR #869, opened 2019-02-15
tkt-75468: Require --force (-f) to start a jail when doing exec/console

### LINKED ISSUE #856: Don't start a jail on calling "iocage console" if it is not running already
Hey guys,

minor remark: similarly to this, iocage should IMHO *not* silently start a jail that is offline when I issue `iocage console foo` ...
Suggestion: error message "jail is down, use -f to force-start" or similar.

Kind regards,
Patrick

_Originally posted by @pmhausen in https://github.com/iocage/iocage/pull/831#issuecomment-460938232_

skarekrow:
Restores old behavior before the commands were folded into the lib.

Closes #856
FreeNAS Ticket: #75468

### DIFF
--- a/iocage_cli/console.py
+++ b/iocage_cli/console.py
@@ -30,12 +30,20 @@ import iocage_lib.iocage as ioc
 __rootcmd__ = True
 
 
-@click.command(name="console", help="Login to a jail.")
-@click.argument("jail")
-def cli(jail):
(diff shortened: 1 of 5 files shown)


sonicaj (dismissed):

sonicaj:
@skarekrow I think the tests need updating as well for to reflect this there

PR #878, opened 2019-02-26
tkt-77421: No longer start a jail when grabbing a plugin property unless forced

skarekrow:
Corrects behavior to be like exec/console
FreeNAS Ticket: #77421


sonicaj (approved):

skarekrow:
Unrelated failure

PR #880, opened 2019-02-26
tkt-77377: Treat stdout being closed as not a tty

skarekrow:
FreeNAS Ticket: #77377


sonicaj (dismissed):
LGTM but do you think it might be better if we perhaps move this to ioc_common ?

skarekrow:
> LGTM but do you think it might be better if we perhaps move this to ioc_common ?

Great idea!

sonicaj (dismissed):

sonicaj (approved):

PR #881, opened 2019-02-27
Improve rctl/cpuset validation

sonicaj:
1) Allow suffixes in rctl props which deal with bytes
2) Make cpuset validation error more informative

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -206,8 +206,23 @@ class IOCCpuset(object):
     def validate_cpuset_prop(value, raise_error=True):
         failed = False
         cpu_sets = IOCCpuset.retrieve_cpu_sets() + 1
+        cpu_set_value_err = False
 
         if not any(
+            cond for cond in (
+                re.findall(
+                    r'^(\d+(,\d+)*)$',
+                    value
+                ),
+                value in ('off', 'all'),
+                re.findall(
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_json.py
```diff
@@ -206,35 +206,54 @@ def retrieve_cpu_sets():
...
+            message = 'Please specify a valid format for cpuset ' \
+                      'value.\nFollowing 4 formats are supported:\n' \
+                      '1) comma delimited string i.e 0,1,2,3\n' \
+                      '2) a range of values i.e 0-2\n' \
+                      '3) "all" - all would mean using all cores\n' \
+                      '4) off'
+            if cpu_set_value_err:
+                message = 'Please make sure the provided cpus fall in the ' \
+                          'correct range of cpus.\nValid cpu entries can be ' \
+                          'found by "cpuset -g -s 0"'
```
skarekrow: Since we know it's now failed, let's store that command in a variable and show the values to the user, saves them a step

skarekrow (changes requested):

iocage_lib/ioc_json.py
sonicaj: @skarekrow this is a good idea ;)

skarekrow (dismissed):

skarekrow (dismissed):

skarekrow (approved):

PR #882, opened 2019-02-28
tkt-77574: Add assign_localhost prop

skarekrow:
This property will assign a localhost of 127.0.1.1 to all jails that have this property enabled. It will also change the /etc/hosts entry to reflect this change if you set it during creation, otherwise it is manual.

- Also allow True/False for truthy props.

FreeNAS Ticket: #77574

### DIFF
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -906,7 +906,7 @@ def get_jail_freebsd_version(path, release):
 
 def check_truthy(value):
     """Checks if the given value is 'True'"""
-    if str(value).lower() in ('1', 'on', 'yes', 'true', 'True'):
+    if str(value).lower() in ('1', 'on', 'yes', 'true'):
         return 1
 
     return 0
@@ -928,3 +928,7 @@ def is_tty():
         is_tty = False
 
     return is_tty
+
+
+def lowercase_set(values):
+    return set([v.lower() for v in values])
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -906,7 +906,7 @@ def get_jail_freebsd_version(path, release):
 
 def check_truthy(value):
     """Checks if the given value is 'True'"""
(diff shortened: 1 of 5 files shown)


iocage_lib/ioc_common.py
```diff
@@ -906,7 +906,7 @@ def get_jail_freebsd_version(path, release):
 
 def check_truthy(value):
     """Checks if the given value is 'True'"""
-    if str(value) in ('1', 'on', 'yes', 'true'):
+    if str(value).lower() in ('1', 'on', 'yes', 'true', 'True'):
```
sonicaj: If we convert it into `lower`, `True` is not needed ;)

iocage_lib/ioc_json.py
```diff
@@ -1018,6 +1018,10 @@ def check_config(self, conf, default=False):
         if not conf.get('exec_created'):
             conf['exec_created'] = '/usr/bin/true'
 
+        # Version 21 keys
```
sonicaj, line 1021: version bump ;)

iocage_lib/ioc_start.py
```diff
@@ -298,6 +299,14 @@ def __start_jail__(self):
 
                 net.append(f'ip6.addr={ip6_addr}')
 
+            if assign_localhost:
+                # Make sure this exists, jail(8) will tear it down if we don't
+                # manually do this.
+                if self.check_aliases('127.0.1.1', '4') != '127.0.1.1':
+                    su.run(['ifconfig', 'lo0', 'alias', '127.0.1.1/32'])
+
+                net.append(f'ip4.addr=127.0.1.1')
```
sonicaj: We would be writing `ip4.addr` twice i think, is that desirable ?

skarekrow: Yes that's fine, jail will only apply it once

sonicaj (approved):

skarekrow:
Unrelated failure

PR #884, opened 2019-03-01
tkt-77574: Add property localhost_ip

skarekrow:
This property will allow a user to specify which address to assign during creation if assign_localhost is True and vnet is False.

- Add new method to generate unused IP's
- Alias localhost IP's during creation
- Check for active jail IP's and warn the user if they have a jail already using this IP as it may be fatal

FreeNAS Ticket: #77574

### DIFF
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -936,29 +936,23 @@ def lowercase_set(values):
     return set([v.lower() for v in values])
 
 
-def generate_unused_ip(ip_prefix, interface='lo0'):
-    """Best effort to try to allocate a private IP for a jail"""
-    interface_addrs = netifaces.ifaddresses(interface)
-    addresses = [ip['addr'] for ips in interface_addrs.values() for ip in ips
-                 if ip['addr'].startswith(ip_prefix)]
+def gen_unused_lo_ip():
+    """Best effort to try to allocate a localhost IP for a jail"""
+    interface_addrs = netifaces.ifaddresses('lo0')
+    inuse = [ip['addr'] for ips in interface_addrs.values() for ip in ips
+             if ip['addr'].startswith('127.0')]
 
-    for ip in addresses:
+    for ip in ipaddress.IPv4Network('127.0.0.0/8'):
         ip = ipaddress.ip_address(ip) + 1
 
-        if 'lo' in interface and not ip.is_loopback:
-            continue
-        elif 'lo' in interface and not str(ip).startswith('127.'):
-            logit(
-                {
-                    'level': 'EXCEPTION',
-                    'message': f'IP: {ip} is not a loopback address.\n'
-                    'If you wish to use a non-RFC5735 compliant address,'
-                    ' please manually set the localhost_ip property.'
-                }
-            )
-        else:
-            if not ip.is_private:
-                continue
-
-        if str(ip) not in addresses:
+        if str(ip) not in inuse:
             return str(ip)
+
+    logit(
+        {
+            'level': 'EXCEPTION',
+            'message': 'An unused RFC5735 compliant localhost address could'
+            ' not be allocated.\nIf you wish to use a non-RFC5735 compliant'
+            ' address, please manually set the localhost_ip property.'
+        }
+    )
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -937,13 +937,28 @@ def lowercase_set(values):
 
 
 def generate_unused_ip(ip_prefix, interface='lo0'):
-    """Best effort to try to allocate an IP for a jail"""
+    """Best effort to try to allocate a private IP for a jail"""
     interface_addrs = netifaces.ifaddresses(interface)
     addresses = [ip['addr'] for ips in interface_addrs.values() for ip in ips
                  if ip['addr'].startswith(ip_prefix)]
 
     for ip in addresses:
-        ip = str(ipaddress.ip_address(ip) + 1)
+        ip = ipaddress.ip_address(ip) + 1
 
-        if ip not in addresses:
-            return ip
+        if 'lo' in interface and not ip.is_loopback:
+            continue
+        elif 'lo' in interface and not str(ip).startswith('127.'):
+            logit(
+                {
+                    'level': 'EXCEPTION',
+                    'message': f'IP: {ip} is not a loopback address.\n'
+                    'If you wish to use a non-RFC5735 compliant address,'
+                    ' please manually set the localhost_ip property.'
+                }
+            )
+        else:
+            if not ip.is_private:
+                continue
+
+        if str(ip) not in addresses:
+            return str(ip)
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -940,10 +940,11 @@ def gen_unused_lo_ip():
     """Best effort to try to allocate a localhost IP for a jail"""
     interface_addrs = netifaces.ifaddresses('lo0')
     inuse = [ip['addr'] for ips in interface_addrs.values() for ip in ips
-             if ip['addr'].startswith('127.0')]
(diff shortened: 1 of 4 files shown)


iocage_lib/ioc_create.py
```diff
@@ -528,8 +562,26 @@ def _create_jail(self, jail_uuid, location):
...
+                                l_ip = config.get('localhost_ip', 'none')
+                                l_ip = l_ip if l_ip != 'none' else \
+                                    iocage_lib.ioc_common.generate_unused_ip(
+                                        '127.')
+                                config['localhost_ip'] = l_ip
+                                iocjson.json_write(config)
+
+                                # If they are creating multiple jails, we want
+                                # this aliased before starting the  jail
+                                su.run(
```
sonicaj, line 576: I think this is not necessary ? We should perhaps only take care of this section when we start a jail. This is not required until the jail actually starts ? And after a reboot - this will cause further issues wrt the ip's not being there at all ( we do handle them when we start a jail and i think that is the best way to do these computations ) ?

iocage_lib/ioc_create.py
```diff
@@ -461,6 +462,39 @@ def _create_jail(self, jail_uuid, location):
...
+                        pass
+
+                    try:
+                        del config['localhost_ip']
+                    except KeyError:
+                        # They may not have specified this
+                        pass
+                else:
+                    config['assign_localhost'] = 0
+                    config['localhost_ip'] = 0
```
sonicaj, line 497: We should add validation for `localhost_ip` prop so that we make sure we only save valid values

iocage_lib/ioc_start.py
```diff
@@ -159,6 +159,7 @@ def __start_jail__(self):
         debug_mode = True if os.environ.get(
             'IOCAGE_DEBUG', 'FALSE') == 'TRUE' else False
         assign_localhost = self.conf['assign_localhost']
+        localhost_ip = self.conf['localhost_ip']
```
sonicaj, line 162: Also i am not sure if this is good, do you think it would make more sense if we only save the value of `localhost_ip` in two cases
1) The user supplies it i.e 127.0.0.3
2) The default value - none

In all other cases we don't reflect this in it's configuration but automatically generate it when the jail starts. 
We can log it though so we are able to retrieve it for debugging ?
Thoughts ?

iocage_lib/ioc_common.py
```diff
@@ -932,3 +934,16 @@ def is_tty():
...
+
+
+def generate_unused_ip(ip_prefix, interface='lo0'):
+    """Best effort to try to allocate an IP for a jail"""
+    interface_addrs = netifaces.ifaddresses(interface)
+    addresses = [ip['addr'] for ips in interface_addrs.values() for ip in ips
+                 if ip['addr'].startswith(ip_prefix)]
+
+    for ip in addresses:
+        ip = str(ipaddress.ip_address(ip) + 1)
```
sonicaj: `127.255.255.255 + 1 = 128.0.0.0` ;)

iocage_lib/ioc_create.py
skarekrow: We actually need to do this at creation, as it allows the alias to then exist so we don't create another jail that uses the same address accidentally.

iocage_lib/ioc_start.py
skarekrow: Automatically generating that is problematic. There are a few issues with that. We'd have to change the jails hosts file each time it boots, which could impact services they have listening for/on these IP's. The safest approach is to generate it once, do our best to ensure uniqueness on the host at that time, and set the property for them to change later. We should avoid changing things inside the jail as often as we can, it would break POLA otherwise.

iocage_lib/ioc_create.py
sonicaj: What happens on a reboot in this case ?

skarekrow: The jail's property will have stored that address for this exact reason :) We helped to ensure the uniqueness at creation to the best of our ability. So if all goes well, they each have a non-conflicting address and come right on up. But that doesn't solve the issue of a new jail being created when those previous jails aren't running, there will be collision there.

iocage_lib/ioc_common.py
```diff
@@ -932,3 +934,31 @@ def is_tty():
...
+    interface_addrs = netifaces.ifaddresses(interface)
+    addresses = [ip['addr'] for ips in interface_addrs.values() for ip in ips
+                 if ip['addr'].startswith(ip_prefix)]
+
+    for ip in addresses:
+        ip = ipaddress.ip_address(ip) + 1
+
+        if 'lo' in interface and not ip.is_loopback:
+            continue
+        elif 'lo' in interface and not str(ip).startswith('127.'):
```
sonicaj: I think this condition will never be executed ?

sonicaj: Also i have a weird case - user has `127.255.255.255` and that's it :D This will return none then.

iocage_lib/ioc_create.py
```diff
@@ -528,8 +562,26 @@ def _create_jail(self, jail_uuid, location):
...
+                                        '127.')
+                                config['localhost_ip'] = l_ip
+                                iocjson.json_write(config)
+
+                                # If they are creating multiple jails, we want
+                                # this aliased before starting the  jail
+                                su.run(
+                                    [
+                                        'ifconfig', 'lo0', 'alias',
+                                        f'{l_ip}/32'
```
sonicaj, line 579: none will error out here

iocage_lib/ioc_json.py
```diff
@@ -2452,6 +2458,18 @@ def json_check_prop(self, key, value, conf):
                     IOCRCTL.validate_rctl_props(key, value)
                 elif key == 'cpuset':
                     IOCCpuset.validate_cpuset_prop(value)
+                elif key == 'localhost_ip':
```
sonicaj, line 2461: I think we need to account for `none` here ? Else we will raise an error on that one too

iocage_lib/ioc_start.py
```diff
@@ -292,13 +293,35 @@ def __start_jail__(self):
...
                 # manually do this.
-                if self.check_aliases('127.0.1.1', '4') != '127.0.1.1':
-                    su.run(['ifconfig', 'lo0', 'alias', '127.0.1.1/32'])
+                if localhost_ip == 'none':
+                    localhost_ip = iocage_lib.ioc_common.generate_unused_ip(
+                        '127.')
+                    self.conf['localhost_ip'] = localhost_ip
+
+                if self.check_aliases(localhost_ip, '4') != localhost_ip:
+                    su.run(['ifconfig', 'lo0', 'alias', f'{localhost_ip}/32'])
```
sonicaj, line 313: This will error out if the value is none

iocage_lib/ioc_start.py
```diff
@@ -292,13 +293,35 @@ def __start_jail__(self):
...
+                            "level": "WARNING",
+                            "message": f'  {self.uuid} is reusing a localhost'
+                                       ' address, failure may occur!'
+                        },
+                            _callback=self.callback,
+                            silent=self.silent)
 
                 if ip4_addr == 'none':
-                    ip4_addr = '127.0.1.1'
+                    ip4_addr = localhost_ip
```
sonicaj, line 333: none will have a fallout here as well

sonicaj (commented):
Thank you ;) Just a few minor notes

iocage_lib/ioc_common.py
skarekrow: This condition should be executed when the user has 128.* (a ton of localhost shared IP jails)

iocage_lib/ioc_json.py
skarekrow: None is an error in this case, since it shouldn't be specified otherwise

sonicaj (approved):

skarekrow:
Tests won't cover this

PR #886, opened 2019-03-04
Change default open to io.open to handle unicode in freebsd-version

pludi:
The file `bin/freebsd-version` in each Jail contains a maintainer-line, which for 12.0 is
```
# Copyright (c) 2013 Dag-Erling Smørgrav
```
which contains a Unicode character that trips the `startswith` in https://github.com/iocage/iocage/blob/1e5339376f376945c365c7d2e5be045aee4b0ff9/iocage_lib/ioc_common.py#L898-L902

This fix changes `open` to `io.open` including setting the correct character set, allowing the code to finish without exception.

### DIFF
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -896,7 +896,9 @@ def get_jail_freebsd_version(path, release):
         # 9.3-RELEASE and under don't actually have this binary
         new_release = release
     else:
-        with io.open(f'{path}/bin/freebsd-version', mode='r', encoding='utf-8') as r:
+        with io.open(
+            f'{path}/bin/freebsd-version', mode='r', encoding='utf-8'
+        ) as r:
             for line in r:
                 if line.startswith('USERLAND_VERSION'):
                     new_release = line.rstrip().partition('=')[
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -37,7 +37,6 @@ import datetime as dt
 import re
 import shlex
 import glob
-import io
 
 import iocage_lib.ioc_exceptions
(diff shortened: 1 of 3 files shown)


skarekrow:
Thanks for this! There's more usages of `freebsd-version` in the program (including some variables named `freebsd_version`), so we should tackle all of those

skarekrow:
Also as far as I know, `io.open` is the same as `open`, so we can drop an import line and just specify the encoding in the open statement?

pludi:
> Also as far as I know, `io.open` is the same as `open`, so we can drop an import line and just specify the encoding in the open statement?

True. Changed that.

> There's more usages of freebsd-version in the program (including some variables named freebsd_version), so we should tackle all of those

I found one other in `ioc_json.py`. The others are checked either through `ioc_common.get_jail_freebsd_version` (which I changed) or `ioc_common.checkoutput`, which already decodes to UTF-8.

sonicaj (commented):
You missed two usages in
1)https://github.com/iocage/iocage/blob/a09857be4f3ef411d0ffd08d4b49fb7267ae7cf8/iocage_lib/ioc_plugin.py#L308
2) https://github.com/iocage/iocage/blob/a09857be4f3ef411d0ffd08d4b49fb7267ae7cf8/iocage_lib/ioc_plugin.py#L327

Can you please have them covered as well ? Thank you ; )

pludi:
True. Changed them & merged current `master` as well.

sonicaj (approved):
Perfect, thank you! Let's wait for the tests ; )

pludi:
Tests are green.

sonicaj:
Thank you ; )

PR #887, opened 2019-03-05
tkt-78546: Don't die on plugin upgrade if snapshot exists

skarekrow:
- Add new class of exception, Exists
- Remove env pollution as upstream has fixed this issue and will cause issues in the FreeNAS middleware
- Raise the new exception on snapshotting when appropriate

FreeNAS Ticket: #78546

### DIFF
--- a/iocage_lib/ioc_plugin.py
+++ b/iocage_lib/ioc_plugin.py
@@ -1038,7 +1038,12 @@ fingerprint: {fingerprint}
             _callback=self.callback,
             silent=self.silent)
 
-        self.__snapshot_jail__(name="upgrade")
+        try:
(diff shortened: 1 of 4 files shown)


iocage_lib/ioc_plugin.py
```diff
@@ -1038,7 +1038,12 @@ def upgrade(self, jid):
             _callback=self.callback,
             silent=self.silent)
 
-        self.__snapshot_jail__(name="upgrade")
+        try:
```
sonicaj, line 1046: Would we like to do this on `update` as well ?

sonicaj (approved):

skarekrow:
Unrelated failure

PR #888, opened 2019-03-08
tkt-76864: Split on tabs for fstab 

skarekrow:
Otherwise we end up getting false validation and mangling mounts with spaces in them.

FreeNAS Ticket: #76864


sonicaj (approved):

skarekrow:
Unrelated failure

PR #897, opened 2019-03-26
tkt-80697: Add NAT support to iocage

skarekrow:
This supports both ipfw and pf backends, iocage will take over the firewall backend you choose, so make sure you aren't relying on it as a firewall.

Usage:
iocage set nat=1 JAIL

You can supply nat_forwards in the following configurations:

PORT <-- Will be assumed to be TCP
PROTO(PORT)
PROTO(PORT:EXTERNALPORT)
PROTO(PORTSTART-PORTEND) <-- Including pf, this will be internally replaced to the correct syntax, as otherwise it will conflict with external portmapping

In this commit:
- Common method to grab all used IP's
- Common method to generate nat IP's
- NAT support for all iocage network operations (pkglist, plugins, etc)
- Fix clean -j not correctly stopping jails
- Add new properties: nat, nat_prefix, nat_interface, nat_backend and nat_forwards
- Add NAT tests and new --nat flag for it
- Fix tests with pkg fetching and lack of allow_raw_sockets
- Use the correct exception for NotImplementedError

Ticket: #80697

### DIFF
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -38,7 +38,8 @@ import re
 import shlex
 import glob
 import netifaces
-import ipaddress
+import concurrent.futures
+import json
 
 import iocage_lib.ioc_exceptions
 import iocage_lib.ioc_exec
@@ -943,11 +944,13 @@ def gen_unused_lo_ip():
              if ip['addr'].startswith('127')]
 
     for ip in ipaddress.IPv4Network('127.0.0.0/8'):
-        if str(ip) == '127.0.0.0':
+        ip_exploded = ip.exploded
+
+        if ip_exploded == '127.0.0.0':
             continue
 
-        if str(ip) not in inuse:
-            return str(ip)
+        if ip_exploded not in inuse:
+            return ip_exploded
 
     logit(
         {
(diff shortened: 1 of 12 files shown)


iocage_lib/ioc_common.py
```diff
@@ -957,3 +960,62 @@ def gen_unused_lo_ip():
...
+    inuse = get_used_ips()
+
+    for i in range(256):
+        for l in range(1, 256, 4):
+            network = ipaddress.IPv4Network(
+                f'{ip_prefix}.{i}.{l}/30', strict=False
+            )
+            pair = [_ip.exploded for _ip in network.hosts()]
+
+            if any(x in pair for x in inuse):
```
sonicaj, line 976: Not necessary - just a suggestion - Using set intersection ( & ) might give us more speed here

iocage_lib/ioc_json.py
```diff
@@ -2292,7 +2310,12 @@ def json_check_prop(self, key, value, conf):
             'rtsold': truth_variations,
             'ip_hostname': truth_variations,
             'assign_localhost': truth_variations,
-            'localhost_ip': ('string', )
+            'localhost_ip': ('string', ),
+            'nat': truth_variations,
+            'nat_prefix': ('string', ),
```
sonicaj, line 2318: We should probably add validation for `nat_prefix`

sonicaj (commented):
This is awesome! Good work @skarekrow ;)
I just have a couple of minor suggestions.

iocage_lib/ioc_common.py
skarekrow: That may be less pythonic and clear though, so I went with the clearer approach

iocage_lib/ioc_json.py
```diff
@@ -2458,7 +2492,7 @@ def json_check_prop(self, key, value, conf):
                     IOCRCTL.validate_rctl_props(key, value)
                 elif key == 'cpuset':
                     IOCCpuset.validate_cpuset_prop(value)
-                elif key == 'localhost_ip':
+                elif key in ('localhost_ip'):
```
william-gr: This is not right, you need a comma to not make it a singleton

iocage_lib/ioc_start.py
```diff
@@ -1245,3 +1305,160 @@ def find_bridge_mtu(self, bridge):
...
+        su.run(
+            ['sysctl', '-q', 'net.inet.ip.fw.enable=1'], capture_output=True
+        )
+        su.run(['kldload', '-n', 'ipfw'])
+
+    def __add_nat__(self, nat_interface, forwards, backend='ipfw'):
+        if backend == 'pf':
+            self.__add_nat_pf__(nat_interface, forwards)
+            pf = su.run(
+                ['pfctl', '-f', '/tmp/iocage_nat_pf.conf'], capture_output=True
```
william-gr: you should probably return `/tmp/iocage_nat_pf.conf` as a value from `__add_nat_pf__` to not repeat yourself. Same for ipfw

sonicaj (dismissed):
Tests are green ;)

skarekrow:
@sonicaj We should be all good (I am locally). Can you confirm and mark as passed?

iocage_lib/ioc_json.py
```diff
@@ -2471,6 +2508,50 @@ def json_check_prop(self, key, value, conf):
...
+                                    silent=self.silent,
+                                    exception=ioc_exceptions.ValidationFailed
+                                )
+
+                        if new_value:
+                            value = ','.join(new_value)
+                            conf[key] = value
+                elif key == 'nat_prefix':
+                    if value != 'none':
+                        if not re.match(r'\d{1,3}\.\d{1,3}$', value):
```
sonicaj: This will allow all values like `999.999` too, i don't think that is desirable ?

sonicaj (approved):
Thank you ;)

PR #899, opened 2019-04-02
tkt-39653: Correctly fetch plugin supplied even if INDEX doesn't exist

skarekrow:
Also default to vnet=1 for plugins

FreeNAS Ticket: #39653


sonicaj (approved):

skarekrow:
Unrelated failure

PR #906, opened 2019-04-04
We only need to split the first 2 commas. If we split all commas and …

SumGuyV5:
…the user passes a string value with one or more commas, they will be removed by the split command and never passed on to the plug-ins set script.

I was trying to make a plug-in for Drupal8 that has a settings were the user passes a string with commas in it to be put in a php array. But I found that the commas were disappearing IE
sudo iocage set -P drupaltrust="'^testplugin$',
  '^localhost$',
  '^192.168.1.50$',
  '^127\.0\.0\.1$'" testplugin

iocage would output
Key: drupaltrust has been updated to '^testplugin$',
  '^localhost$',
  '^192.168.1.50$',
  '^127\.0\.0\.1$'

But when I checked into the Drupal settings.php I found the commas were gone.

When I go into the jail and run /usr/local/bin/drupal8set drupaltrust "'^testplugin$',
  '^localhost$',
  '^192.168.1.50$',
  '^127\.0\.0\.1$'"

Things worked fine.

So I add a echo $@ to my script to see what was being passed by iocage and found the string was missing the commas.

I had this issue on 
Version 1.0 RELEASE 2018/12 
and
Version FreeNAS 11.2-U3

This little change should fix the issue.

- ["CHECK ] Explain the feature
- ["CHECK" ] Read [CONTRIBUTING.md](https://github.com/iocage/iocage/blob/master/CONTRIBUTING.md)


jsegaert:
This may break the ability to pass in more than one argument to the _pluginset_ script though; this is used in e.g. the quassel plugin.  Also see [this note](https://www.ixsystems.com/community/threads/purpose-of-settings-json-in-iocage-plugins.75268/#post-523362) for another plugin in progress.
@skarekrow Maybe a more flexible solution is in order?

skarekrow:
Hmm, @jsegaert is right. @SumGuyV5 does escaping the commas help in this situation?

skarekrow:

PR #907, opened 2019-04-04
Use .replace('.', '_') here as well?

kcbark:
I believe we should replace . with _ here as well as we do in list_get_jid ioc_list.py?
This unborked my jails (named www_test_com) after upgrading from iocage 1.1 to master.

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -1536,7 +1536,7 @@ class IOCJson(IOCConfiguration):
         if state:
             ruleset = su.check_output(
                 [
-                    "jls", "-j", f"ioc-{conf['host_hostuuid'].replace('.', '_')}",
+                    "jls", "-j", f'ioc-{conf["host_hostuuid"].replace(".", "_")}',
                     'devfs_ruleset'
                 ]
             ).decode().rstrip()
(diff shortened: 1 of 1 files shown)


skarekrow (changes requested):
Yes we should, let's keep the single quotes though :P

iocage_lib/ioc_json.py
```diff
@@ -1536,7 +1536,7 @@ def get_full_config(self):
         if state:
             ruleset = su.check_output(
                 [
-                    'jls', '-j', f'ioc-{conf["host_hostuuid"]}',
+                    "jls", "-j", f'ioc-{conf["host_hostuuid"].replace(".", "_")}',
```
skarekrow:
```suggestion
                    'jls', '-j', f'ioc-{conf["host_hostuuid"].replace(".", "_")}',
```

skarekrow (dismissed):

skarekrow (dismissed):

skarekrow:
Thanks @kcbark!

skarekrow:
@kcbark The travis test will need to be fixed though before merging :)

kcbark:
Ah, didn’t see that. I’ll fix :)

skarekrow (approved):

PR #909, opened 2019-04-05
Fix fstabs (for real) with spaces and other bad characters

skarekrow:
We now use strvis and strunvis to correctly encode/decode these the way fstab would like.

NAS-101252

### DIFF
--- a/iocage_lib/ioc_fstab.py
+++ b/iocage_lib/ioc_fstab.py
@@ -407,7 +407,7 @@ class IOCFstab(object):
 
         Example: ' ' -> \040
         """
-        result = ctypes.create_string_buffer(1024)
+        result = ctypes.create_string_buffer(len(_string) * 4 + 1)
         self.libc.strvis(
             result, _string.encode(), 0x4 | 0x8 | 0x10 | 0x2000 | 0x8000
         )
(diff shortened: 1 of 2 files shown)


iocage_lib/ioc_fstab.py
```diff
@@ -378,3 +400,29 @@ def fstab_list(self):
...
 
         return table.draw()
+
+    def __fstab_encode__(self, _string):
+        """
+        Will parse any non-acceptable fstab characters and encode them
+
+        Example: ' ' -> \040
+        """
+        result = ctypes.create_string_buffer(1024)
```
themylogin: Are you sure it will always be enough?

skarekrow: That's fstab's max length, so it should be (https://github.com/freebsd/freebsd/blob/master/lib/libc/gen/fstab.c#L114)

themylogin: `The size of dst must be four times the number of characters encoded from src (plus one for the NUL).` - from the manpage. It'll be much safer if we use this calculation instead of `1024`.

themylogin: @skarekrow Yes, but value passed to `__fstab_encode__` does not come from fstab and it will cause segmentation fault if it won't fit.

themylogin (approved):
Thank you!

PR #921, opened 2019-04-23
NAS-101365 / 11.3 / Proxy support for iocage

sonicaj:
This commit adds proxy support for iocage making sure that proxy provided by cli or set in the environment if library is being used directly is respected.
Ticket: NAS-101365

Note: DNS lookups DONOT respect the proxy, so a proper firewall should be configured - other communications, fetching, installing packages etc obey the proxy as desired

### DIFF
--- a/iocage_cli/create.py
+++ b/iocage_cli/create.py
@@ -52,45 +52,73 @@ def validate_count(ctx, param, value):
         return int(value)
 
 
-@click.command(name="create", help="Create a jail.")
-@click.option("--count", "-c", callback=validate_count, default="1",
-              help="Designate a number of jails to create. Jails are"
-                   " numbered sequentially.")
-@click.option("--thickconfig", "-C", default=False, is_flag=True,
-              help="Do not use inheritable configuration with jails")
-@click.option("--release", "-r", required=False,
-              help="Specify the RELEASE to use for the new jail.")
-@click.option("--template", "-t", required=False,
-              help="Specify the template to use for the new jail instead of"
-                   " a RELEASE.")
-@click.option("--pkglist", "-p", default=None,
-              help="Specify a JSON file which manages the installation of"
-                   " each package in the newly created jail.")
-@click.option("--name", "-n", default=None,
-              help="Provide a specific name instead of an UUID for this jail.")
-@click.option("--uuid", "-u", "_uuid", default=None,
-              help="Provide a specific UUID for this jail.")
-@click.option("--basejail", "-b", is_flag=True, default=False,
-              help="Set the new jail type to a basejail. Basejails are"
-                   " thick jails (unless specified) that mount the specified"
-                   " RELEASE directories as nullfs mounts over the jail's"
-                   " directories.")
-@click.option("--clone_basejail", "-B", is_flag=True, default=False,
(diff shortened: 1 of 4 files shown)


iocage_cli/create.py
```diff
@@ -52,45 +52,73 @@ def validate_count(ctx, param, value):
...
+@click.option(
+    "--name", "-n", default=None,
+    help="Provide a specific name instead of an UUID for this jail."
+)
+@click.option(
+    "--uuid", "-u", "_uuid", default=None,
+    help="Provide a specific UUID for this jail."
+)
+@click.option(
+    '--proxy', '-S', default=None,
```
skarekrow, line 89: Why `-S`?

iocage_lib/ioc_create.py
```diff
@@ -879,7 +879,9 @@ def create_install_packages(self, jail_uuid, location,
                     _callback=self.callback)
 
         # We will have mismatched ABI errors from earlier, this is to be safe.
-        pkg_env = {"ASSUME_ALWAYS_YES": "yes"}
+        pkg_env = {
+            **os.environ.copy(), "ASSUME_ALWAYS_YES": "yes"
```
skarekrow: This will leak the users environment during execution (ssh and such), we should keep this minimal and add just the two proxy variables.

iocage_lib/ioc_plugin.py
```diff
@@ -564,7 +564,9 @@ def __fetch_plugin_post_install__(self, conf, _conf, jaildir, uuid):
 
             os.environ["IOCAGE_PLUGIN_IP"] = ip
 
-        plugin_env = {"IOCAGE_PLUGIN_IP": ip.rsplit(',')[0]}
+        plugin_env = {
+            **os.environ.copy(), 'IOCAGE_PLUGIN_IP': ip.rsplit(',')[0]
```
skarekrow: Same here

iocage_lib/ioc_plugin.py
```diff
@@ -1331,21 +1333,35 @@ def __clone_repo(self, repo_url, destination):
...
             origin = repo.remotes.origin
-        except git.exc.InvalidGitRepositoryError:
+            ref = 'master' if f'origin/{ref}' not in repo.refs else ref
+            for command in [
+                ['checkout', ref],
+                ['pull']
+            ]:
+                su.check_call(
+                    ['git', '-C', destination] + command,
+                    stdout=su.PIPE, stderr=su.PIPE, env=os.environ.copy()
```
skarekrow: Same here

skarekrow (changes requested):
This needs to have the proxy support added to lib/iocage so we can use it reliably from the CLI and FreeNAS.

Currently this only supports CLI usage.

skarekrow (dismissed):

skarekrow (approved):

tq02ksu:
wonderful feature! but why this can not be used with freeNAS web-UI? Might be the reason that need freenas web-ui update invocation logic to fit the new parameters?

sonicaj:
@tq02ksu  it can be used with FreeNAS, but is available in latest 11.3 version

tq02ksu:
@sonicaj Thanks for you infomation!

PR #923, opened 2019-04-24
NAS-101531 / 11.3 / Do not test len of None

skarekrow:
NAS-101531

Signed-off-by: Brandon Schneider <brandon@ixsystems.com>

### DIFF
--- a/iocage_lib/ioc_fstab.py
+++ b/iocage_lib/ioc_fstab.py
@@ -463,6 +463,9 @@ class IOCFstab(object):
 
         Example: ' ' -> \040
         """
+        if _string is None:
+            return _string
+
         result = ctypes.create_string_buffer(len(_string) * 4 + 1)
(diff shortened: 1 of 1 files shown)


sonicaj:
So i think this requires changes in the jail plugin on FN end as well to ensure that we have defaults for source/destination in case index is specified - also i think we can improve the validation there by ensuring that if index is not provided - we ensure other values are provided for actions other then list.

skarekrow:
> So i think this requires changes in the jail plugin on FN end as well to ensure that we have defaults for source/destination in case index is specified - also i think we can improve the validation there by ensuring that if index is not provided - we ensure other values are provided for actions other then list.

That shouldn't be required as the defaults are handled in the library itself now.

sonicaj:
Nope - we pass None from FN end - also we don't validate properly the two cases i.e with/without index props

PR #927, opened 2019-05-06
NAS-101624 / 11.3 / Fix bad usage of plugin names

skarekrow:
These are artifacts of a time when iocage didn’t support multiple plugins at the same time.

This PR should fix the following behaviors:
- Update
- Upgrade
- Destruction on Exception

NAS-101624


sonicaj (commented):
We should cover https://github.com/iocage/iocage/blob/ce967e2e0caded851ff2300fc1ed5b3b6a15f503/iocage_lib/ioc_plugin.py#L779 this usage as well to reflect `fetch_plugin` changes

skarekrow:
> We should cover
> 
> https://github.com/iocage/iocage/blob/ce967e2e0caded851ff2300fc1ed5b3b6a15f503/iocage_lib/ioc_plugin.py#L779
> this usage as well to reflect `fetch_plugin` changes

What do you mean? That shouldn't require any changes.

sonicaj:
> > We should cover
> > https://github.com/iocage/iocage/blob/ce967e2e0caded851ff2300fc1ed5b3b6a15f503/iocage_lib/ioc_plugin.py#L779
> > 
> > this usage as well to reflect `fetch_plugin` changes
> 
> What do you mean? That shouldn't require any changes.

We only allow 3 parameters now instead of four ;) So this should be updated to reflect that

iocage_lib/ioc_plugin.py
```diff
@@ -149,18 +148,18 @@ def fetch_plugin(self, _json, props, num, accept_license):
...
-                                                   pkg, props, repo_dir)
-            self.__fetch_plugin_post_install__(conf, _conf, jaildir, jail_name)
-        except Exception as e:
+            jaildir, _conf, repo_dir = self.__fetch_plugin_create__(props)
+            self.__fetch_plugin_install_packages__(
+                jaildir, conf, pkg, props, repo_dir
+            )
+            self.__fetch_plugin_post_install__(conf, _conf, jaildir)
+        except BaseException as e:
+            self.keep_jail_on_failure = False
```
sonicaj: I am sorry i am not sure if i follow this 🤔 Why do we hardcode this value here on an exception ?
Do we want to destroy plugins always on failure ?

sonicaj (commented):
Looks good to me ;) Just one minor concern

iocage_lib/ioc_plugin.py
skarekrow: Debug :D I can't believe I forgot it.

sonicaj: Happens to the best of us ;)

sonicaj (approved):
Thank you

PR #928, opened 2019-05-08
NAS-101560 / 11.3 / Warn users for bad perms

sonicaj:
This commit adds a warning for users who have accidentally/deliberately changed ownership/permissions for /usr/lib inside jail. The warning appears when the user tries iocage console, we don't error out but just give the user a friendly notice that there is a good chance login(1) is not going to like the changed perms and so he/she should have that corrected.
Ticket: #NAS-101560

### DIFF
--- a/iocage_lib/iocage.py
+++ b/iocage_lib/iocage.py
@@ -897,7 +897,19 @@ class IOCage(ioc_json.IOCZFS):
             login_flags = self.get('login_flags').split()
             console_cmd = ['login', '-p'] + login_flags
 
-            ioc_exec.InteractiveExec(console_cmd, path, uuid=uuid)
+            try:
+                ioc_exec.InteractiveExec(console_cmd, path, uuid=uuid)
+            except BaseException as e:
+                ioc_common.logit(
+                    {
+                        'level': 'ERROR',
+                        'message': 'Console failed!\nThe cause could be bad '
+                                   f'permissions for {path}/root/usr/lib.'
+                    },
+                    _callback=self.callback,
+                    silent=False
+                )
+                raise e
             return
 
         if interactive:


iocage_cli/console.py
```diff
@@ -41,7 +41,7 @@ def cli(jail, force):
     # Command is empty since this command is hardcoded later on.
     ioc.IOCage(
         jail=jail,
-        silent=True
+        silent=False
```
skarekrow: Let’s keep this True

iocage_lib/iocage.py
```diff
@@ -897,6 +897,52 @@ def exec(self,
...
+                        {
+                            'level': 'ERROR',
+                            'message': '*** Bad ownership / permissions found '
+                                       f'for files in {path}/root/usr/lib/, '
+                                       'this may lead to a failure getting '
+                                       f'console access to {uuid} jail. '
+                                       'Please have these corrected.'
+                        },
+                        _callback=self.callback,
+                        silent=self.silent
```
skarekrow: And just make it False here

skarekrow (changes requested):

iocage_lib/iocage.py
```diff
@@ -897,7 +897,21 @@ def exec(self,
...
             console_cmd = ['login', '-p'] + login_flags
 
-            ioc_exec.InteractiveExec(console_cmd, path, uuid=uuid)
+            try:
+                ioc_exec.InteractiveExec(console_cmd, path, uuid=uuid)
+            except BaseException as e:
+                ioc_common.logit(
+                    {
+                        'level': 'ERROR',
+                        'message': '*** This could have likely occurred '
```
skarekrow: We should be rewording this to something like:
```
f'Console failed!\nThe cause could be bad permissions for {path}/root/usr/lib'
```

iocage_lib/iocage.py
```diff
@@ -897,7 +897,21 @@ def exec(self,
...
+                                   'because of bad permissions for '
+                                   f'{path}/root/usr/lib, please if that is '
+                                   'the case, have this corrected.'
+                    },
+                    _callback=self.callback,
+                    silent=False
+                )
+                raise e
```
skarekrow, line 912: At this point we should raise a CommandFailed, as otherwise you will get an ugly stacktrace on the CLI.

iocage_lib/iocage.py
```diff
@@ -897,7 +897,21 @@ def exec(self,
             login_flags = self.get('login_flags').split()
             console_cmd = ['login', '-p'] + login_flags
 
-            ioc_exec.InteractiveExec(console_cmd, path, uuid=uuid)
+            try:
+                ioc_exec.InteractiveExec(console_cmd, path, uuid=uuid)
+            except BaseException as e:
+                ioc_common.logit(
+                    {
+                        'level': 'ERROR',
```
skarekrow, line 905:
```suggestion
                        'level': 'EXCEPTION',
```

skarekrow (changes requested):

iocage_lib/iocage.py
sonicaj: An exception has already been raised by `InteractiveExec`, should we raise another one in this case ?

iocage_lib/iocage.py
sonicaj: This is just the SystemExit going on and no stacktrace ;) Should we raise `CommandFailed` ?
```
root@bsd11:~ # iocage console jail
Command: login -p -f root failed!
*** Bad ownership / permissions found for files in /zroot/iocage/jails/jail/root/usr/lib/, this likely caused failure getting console access to jail jail. Please have these corrected.
root@bsd11:~ #
```

iocage_lib/iocage.py
skarekrow: Yes otherwise we lose the nice exception and end up with a stacktrace

iocage_lib/iocage.py
skarekrow: How does that look from the API? I may be misremembering :P

sonicaj: Also if do `EXCEPTION` above, we already have initiated `SytemExit`, no point in raising `CommandFailed`, thoughts ?

skarekrow (approved):

PR #929, opened 2019-05-13
NAS-101668 / 11.3 / Split on both spaces and tabs for fstab

skarekrow:
Previous behavior erroneously expected iocage to be the sole consumer of the fstab.

- This now will skip comments, and split on spaces/tabs as it should have before.
- Move reading fstab to fstab_list() method

iocage side of freenas/freenas#3041


sonicaj (approved):

joachimmathes:
It works with our fstab layout! Thanks a lot! :smiley:

PR #930, opened 2019-05-14
Allow case insensitive suffixes for rctl values

joachimmathes:
Since the `rctl` command allows upper and lower case suffixes, iocage should do that, too. (IMHO)

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -360,7 +360,8 @@ class IOCRCTL(object):
             IOCRCTL.validate_rctl_tunable()
 
             if not re.findall(
-                r'(?:deny|log|devctl|sig\w*|throttle)=\d+(?:b|k|m|g|t|p|)$',
+                r'(?:deny|log|devctl|sig\w*|throttle)='
+                r'\d+(?:[Bb]|[Kk]|[Mm]|[Gg]|[Tt]|[Pp]|)$',
                 value
             ):
                 iocage_lib.ioc_common.logit(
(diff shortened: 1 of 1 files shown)


sonicaj (dismissed):

joachimmathes:
Hmm, the tests seem to run into a timeout.

skarekrow:
@joachimmathes Unfortunately they do that a lot. We've tried to fix that in later PR's, so hopefully that problem becomes less of one down the line. I reran them to be sure here.

iocage_lib/ioc_json.py
```diff
@@ -384,8 +386,8 @@ def validate_rctl_props(prop, value):
                     iocage_lib.ioc_common.logit(
                         {
                             'level': 'EXCEPTION',
-                            'message': 'b, k, m, g, t, p suffixes are not '
```
skarekrow, line 387: I'm wondering if we can't just grab the last character of the value and print that here instead of a long list of every possible capitalization.

Ex: value[-1]

skarekrow (changes requested):

iocage_lib/ioc_json.py
joachimmathes: Sounds reasonable.

joachimmathes: Fixed it.

skarekrow (approved):

skarekrow:
@sonicaj You also good with this still?

sonicaj (approved):

iocage_lib/ioc_json.py
```diff
@@ -360,7 +360,8 @@ def validate_rctl_props(prop, value):
             IOCRCTL.validate_rctl_tunable()
 
             if not re.findall(
-                r'(?:deny|log|devctl|sig\w*|throttle)=\d+(?:b|k|m|g|t|p|)$',
+                r'(?:deny|log|devctl|sig\w*|throttle)='
+                r'\d+(?:[Bb]|[Kk]|[Mm]|[Gg]|[Tt]|[Pp]|)$',
                 value
```
igalic, line 365: why not use `flags=re.IGNORECASE`?

igalic (commented):
🙋

iocage_lib/ioc_json.py
sonicaj: Thank you for pointing it out @igalic. It affected actions as well but having a second look confirms that actions are case insensitive. So at the end i think a better approach is doing a lower on the value and keeping the validation as it was originally. Thanks again!

PR #934, opened 2019-05-21
Remove pcpu limit

joachimmathes:
Starting jails with `iocage start` and `pcpu: "deny=200"` in `config.json` was no problem, but `iocage create ... pcpu="deny=200" ...` failed.

The reason is a condition in `ioc_json.py` where `pcpu` values greater than 100 raise an exception. But it is possible to set `pcpu` to a value greater than 100 with `rctl`. And we do that a lot in our shared host environments with multiple cores. The [rctl documentation](https://man.openbsd.org/FreeBSD-12.0/rctl.8) is misleading and actually I couldn't find a better source than this [bug](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=189870) so far.

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -413,15 +413,6 @@ class IOCRCTL(object):
                         }
                     )
 
-                if prop == 'pcpu' and int(value.split('=')[1]) > 100:
-                    iocage_lib.ioc_common.logit(
-                        {
-                            'level': 'EXCEPTION',
-                            'message': 'pcpu property requires a valid '
-                                       'percentage'
-                        }
-                    )
-
 
 class IOCSnapshot(object):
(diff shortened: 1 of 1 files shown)


skarekrow:
LGTM, @sonicaj?

skarekrow (approved):

sonicaj (approved):
LGTM

skarekrow:
(Using this PR to see if the tests still timeout, it's very unlikely to cause failures itself)

joachimmathes:
Thank you for the fast reply. And no timeout this time. :smiley: 

Would be great to have this merged as soon as possible, since this issue blocks our system from creating new jails (which is our daily business). Thanks again!

skarekrow:
All done :) thanks again!

Issue #936, opened 2019-05-23
jail root nullfs won't unmount (device busy)


joachimmathes:
- [x] Supply `iocage --version`
  Version 1.2 RC (`6474acdf65169e4cd6711d7144b9fe574d1c8397`)
- [x] Supply the commands used, along with any steps to recreate it.
  `iocage stop <JAIL>`
- [x] Provide the output from the command you issued.
  ```
  * Stopping <JAIL>
    + Successfully removed RCTL rules for <JAIL>
    + Executing prestop OK
    + Stopping services OK
    + Tearing down VNET OK
    + Removing devfs_ruleset: 8 OK
    + Removing jail process FAILED:
  umount: unmount of /iocage/jails/<JAIL>/root failed: Device busy
  jail: ioc-<JAIL>: /sbin/umount -t nullfs /iocage/jails/<JAIL>/root: failed
  ```
- [x] Supply what you expected the result or output to be
  ```
  * Stopping <JAIL>
    + Successfully removed RCTL rules for <JAIL>
    + Executing prestop OK
    + Stopping services OK
    + Tearing down VNET OK
    + Removing devfs_ruleset: 8 OK
    + Removing jail process OK
    + Executing poststop OK
  ```
- [x] Checked that the problem has not already been fixed on `master` if using
  a stable release.

The jail fstab:
```
zroot/iocage/jails/<JAIL> on /iocage/jails/<JAIL> (zfs, local, nfsv4acls)
/iocage/jails/<JAIL_TEMPLATE>/root on /iocage/jails/<JAIL>/root (nullfs, local, read-only)
zroot/iocage/jails/<JAIL>/root/etc on /iocage/jails/<JAIL>/root/etc (zfs, local, nfsv4acls)
...
```

The problem of unmounting jail directories seems to be an old and odd
one:
* https://forums.freebsd.org/threads/jail-root-nullfs-wont-unmount-device-busy.45806/
* https://forums.freebsd.org/threads/jail-mounted-base-nullfs-sometimes-not-un-mounted-after-jail-stop.69415/

And it is obviously not a problem of `iocage` itself but of the
underlying `jail` command.

Nevertheless, the problem occurred with the introduction of using
`/var/run/jail.ioc-{self.uuid}.conf` files in iocage, i.e. official
release `1.1` https://github.com/iocage/iocage/blob/3e0c61ada9a22e273590e5c579e65928c61c59b3/iocage_lib/ioc_stop.py#L309

[...]

PR #937, opened 2019-05-24
Add safety net for umount root (#936)

joachimmathes:
Replacing the `EXCEPTION` by an `ERROR` would work:

Before:
```
* Stopping <JAIL>
  + Successfully removed RCTL rules for <JAIL>
  + Executing prestop OK
  + Stopping services OK
  + Tearing down VNET OK
  + Removing devfs_ruleset: 8 OK
  + Removing jail process FAILED:
umount: unmount of /iocage/jails/<JAIL>/root failed: Device busy
jail: ioc-<JAIL>: /sbin/umount -t nullfs /iocage/jails/<JAIL>/root: failed
```

After:
```
iocage stop <JAIL>
* Stopping <JAIL>
  + Successfully removed RCTL rules for <JAIL>
  + Executing prestop OK
  + Stopping services OK
  + Tearing down VNET OK
  + Removing devfs_ruleset: 8 OK
  + Removing jail process FAILED:
umount: unmount of /iocage/jails/<JAIL>/root failed: Device busy
jail: ioc-<JAIL>: /sbin/umount -t nullfs /iocage/jails/<JAIL>/root: failed

  + Executing poststop OK
```


iocage_lib/ioc_stop.py
```diff
@@ -338,7 +338,7 @@ def __stop_jail__(self):
             msg = f'  + Removing jail process FAILED:\n' \
                 f'{stop_err.decode("utf-8")}'
             iocage_lib.ioc_common.logit({
-                'level': 'EXCEPTION',
+                'level': 'ERROR',
```
sonicaj, line 341: @skarekrow i don't like this change. A lot of things can go wrong here and we shouldn't be executing `poststop` in those cases. I do however like the case of manually handling the umount commands for the time being. We can add a note specifying why we have reverted to this and in future change it when the upstream has fixed it. Thoughts ?

skarekrow: I can’t recall exactly, but I think we can do the unmounting before removal of the jail process, since all processes should be shut down. That may allow us to keep the exception and introduce the behavior.

joachimmathes: What we do in our iocage-port for now is:
```diff
--- iocage_lib/ioc_stop.py.orig	2019-05-23 17:54:09 UTC
+++ iocage_lib/ioc_stop.py
@@ -322,9 +322,6 @@ class IOCStop(object):
         # will be written on the next jail start/restart.
         jail_conf_file = Path(f"/var/run/jail.ioc-{self.uuid}.conf")
 
-        if jail_conf_file.is_file():
-            cmd.extend(['-f', str(jail_conf_file)])
-
         cmd.extend(['-r', f'ioc-{self.uuid}'])
 
         stop = su.Popen(
```

I know, neither this "pull request" nor this patch is a good solution, but it works for the moment.

skarekrow:

PR #943, opened 2019-06-06
NAS-101507 / 11.3 / Improve jail updates

sonicaj:
This commit introduces following changes:
1) A flag to update pkg repositories in the jail and then upgrade all existing pkgs in the jail.
2) Ability to update jail itself instead of only the software inside them for plugins.
3) Allow running pkg/exec for ALL jails.

Ticket: #NAS-101507

### DIFF
--- a/iocage_cli/update.py
+++ b/iocage_cli/update.py
@@ -29,11 +29,17 @@ __rootcmd__ = True
 
 
 @click.command(
-    name="update",
-    help="Run freebsd-update to update a specified "
-    "jail to the latest patch level.")
-@click.argument("jail", required=True)
-def cli(jail):
+    name='update',
+    help='Run freebsd-update to update a specified '
+    'jail to the latest patch level.'
+)
+@click.option(
+    '--pkgs', '-P', default=False, is_flag=True,
+    help='Decide whether or not to update the pkg repositories and '
+         'all installed packages in jail( this has no effect for plugins ).'
+)
+@click.argument('jail', required=True)
+def cli(jail, pkgs):
     """Update the supplied jail to the latest patchset"""
     skip_jails = bool(jail != 'ALL')
-    ioc.IOCage(jail=jail, skip_jails=skip_jails).update()
+    ioc.IOCage(jail=jail, skip_jails=skip_jails).update(pkgs=pkgs)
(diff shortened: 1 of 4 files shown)


iocage_cli/update.py
```diff
@@ -29,11 +29,27 @@
...
+    help='Run freebsd-update to update a specified '
+    'jail to the latest patch level.'
+)
+@click.option(
+    '--pkgs/--nopkgs', default=False,
+    help='Decide whether or not to update the pkg repositories and '
+         'all installed packages in jail.'
+)
+@click.option(
+    '--plugins/--noplugins', '-P/-NP', default=True,
```
skarekrow: This doesn't make sense, they're calling update, whatever the type is should always be updated.

iocage_cli/update.py
```diff
@@ -29,11 +29,27 @@
...
+    help='Decide whether or not to update the plugin.'
+)
+@click.option(
+    '--jailversion/--nojailversion', '-J/-NJ', default=True,
+    help='Decide whether or not to update the plugin.'
```
skarekrow: Same here

iocage_cli/update.py
```diff
@@ -29,11 +29,27 @@
...
-@click.argument("jail", required=True)
-def cli(jail):
+    name='update',
+    help='Run freebsd-update to update a specified '
+    'jail to the latest patch level.'
+)
+@click.option(
+    '--pkgs/--nopkgs', default=False,
+    help='Decide whether or not to update the pkg repositories and '
+         'all installed packages in jail.'
```
skarekrow: We should make this true if the jail in question is a plugin jail, this flag should only be false for plugins if they specifically pass --nopkgs

skarekrow (changes requested):

iocage_cli/update.py
sonicaj: I am not sure but to be sure, is there potential for conflict between plugin updates and pkg updates in the same jail ?

skarekrow: They're the same thing. Plugin updates currently are only pkg's + snapshot. So if they specifically disable that, all we should do in the plugin update process is snapshot and then update the world.

iocage_cli/update.py
sonicaj: There's a usecase, say latest plugin update version has bugs. So they would only like to update freebsd jail. In that case they are able to do so. We default to `True` so this is executed in most cases.

skarekrow: That isn't necessary, the pkg flag would control that. The jail should always be updated. It's why they're calling it :P

skarekrow (approved):

PR #948, opened 2019-06-12
NAS-102084 / 11.3 / Improve jail configuration checks

sonicaj:
This commit improves jail configuration checks and fixes a bug where we referenced a variable which did not exist.

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -1082,9 +1082,10 @@ class IOCConfiguration(IOCZFS):
         """
         release = conf.get('release', None)
         template = conf.get('template', 0)
+        host_hostuuid = conf.get('host_hostuuid', None)
         renamed = False
 
-        if release is None:
+        if release is None or host_hostuuid is None:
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_json.py
```diff
@@ -1136,20 +1136,17 @@ def check_jail_config(self, conf):
...
+                )
             except Exception as e:
                 iocage_lib.ioc_common.logit(
                     {
                         'level': 'EXCEPTION',
                         'message': 'Exception:'
-                        f" '{e.__class__.__name__}:{str(e)}' occured\n"
-                        f"Loading {uuid}'s configuration failed"
+                        f" '{e.__class__.__name__}:{str(e)}' occurred\n"
+                        f"Loading {str(freebsd_version).split('/')[-4]}'s "
```
skarekrow: This would be `"Loading 11.2-RELEASE's configuration failed"`, which is why we used the UUID as it's the jail

skarekrow (changes requested):

skarekrow (approved):

PR #953, opened 2019-06-17
NAS-100704 / 11.3 / Add method to retrieve plugin versions

sonicaj:
This commit adds the ability to retrieve plugin versions via packagesite defined for each plugin. As packagesite can differ for each plugin, the only reliable way to make sure that version is valid is to use packagesite.txz and extract version from there.

For now this isn't exposed to CLI because i don't see a need to the why here ( time to execute one of the main reasons ). If we get a feature request or in the future, we can easily add a flag to expose this in the CLI.

### DIFF
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -1025,3 +1025,17 @@ def get_used_ips():
                     addresses.append(address)
 
     return addresses
+
+
+def parse_package_name(pkg):
+    pkg, version = pkg.rsplit('-', 1)
+    epoch_split = version.rsplit(',', 1)
+    epoch = epoch_split[1] if len(epoch_split) == 2 else '0'
+    revision_split = epoch_split[0].rsplit('_', 1)
+    revision = \
+        revision_split[1] if len(revision_split) == 2 else '0'
+    return {
+        'version': revision_split[0],
+        'revision': revision,
+        'epoch': epoch,
+    }
--- a/iocage_lib/ioc_plugin.py
+++ b/iocage_lib/ioc_plugin.py
@@ -28,12 +28,12 @@ import datetime
 import distutils.dir_util
 import json
 import os
+import git
 import pathlib
 import re
 import shutil
 import subprocess as su
 import requests
-import git
 
 import iocage_lib.ioc_common
 import iocage_lib.ioc_create
@@ -94,7 +94,10 @@ class IOCPlugin(object):
 
         if os.geteuid() == 0:
             try:
-                self.__clone_repo(git_server, git_working_dir, depth)
+                self._clone_repo(
+                    self.branch, git_server, git_working_dir,
+                    depth, self.callback
+                )
             except Exception as err:
                 iocage_lib.ioc_common.logit(
                     {
@@ -104,7 +107,8 @@ class IOCPlugin(object):
                     _callback=self.callback
                 )
 
-    def fetch_plugin_packagesites(self, package_sites):
+    @staticmethod
+    def fetch_plugin_packagesites(package_sites):
         def download_parse_packagesite(packagesite_url):
             package_site_data = {}
             try:
@@ -134,8 +138,9 @@ class IOCPlugin(object):
 
         return plugin_packagesite_mapping
 
-    def fetch_plugin_versions_from_plugin_index(self, plugins_index):
-        plugin_packagesite_mapping = self.fetch_plugin_packagesites([
+    @staticmethod
+    def fetch_plugin_versions_from_plugin_index(plugins_index):
+        plugin_packagesite_mapping = IOCPlugin.fetch_plugin_packagesites([
             v['packagesite'] for v in plugins_index.values()
         ])
 
@@ -159,11 +164,10 @@ class IOCPlugin(object):
 
         return version_dict
 
-    def fetch_plugin_versions(self):
-        self.clone_repo()
-
+    @staticmethod
+    def retrieve_plugin_index_data(plugin_index_path):
         with open(
-            os.path.join(self.iocroot, '.plugin_index', 'INDEX'), 'r'
+            os.path.join(plugin_index_path, 'INDEX'), 'r'
         ) as f:
             index = json.loads(f.read())
 
@@ -173,11 +177,18 @@ class IOCPlugin(object):
                 'primary_pkg': index[plugin].get('primary_pkg'),
             }
             with open(
-                os.path.join(
-                    self.iocroot, '.plugin_index', index[plugin]['MANIFEST']
-                ), 'r'
+                os.path.join(plugin_index_path, index[plugin]['MANIFEST']), 'r'
             ) as f:
                 plugin_index[plugin].update(json.loads(f.read()))
+
+        return plugin_index
+
+    def fetch_plugin_versions(self):
+        self.clone_repo()
+
+        plugin_index = self.retrieve_plugin_index_data(
+            os.path.join(self.iocroot, '.plugin_index')
+        )
 
         return self.fetch_plugin_versions_from_plugin_index(plugin_index)
 
@@ -674,7 +685,10 @@ fingerprint: {fingerprint}
                 _callback=self.callback,
                 silent=self.silent)
 
-            self.__clone_repo(conf['artifact'], f'{jaildir}/plugin')
+            self._clone_repo(
+                self.branch, conf['artifact'],
+                f'{jaildir}/plugin', callback=self.callback
+            )
 
             with open(
                 f"{jaildir}/{self.jail.rsplit('_', 1)[0]}.json", "w"
@@ -1013,14 +1027,19 @@ fingerprint: {fingerprint}
 
         git_working_dir = f"{self.iocroot}/.plugin_index"
 
-        self.__clone_repo(git_server, git_working_dir)
+        self._clone_repo(
+            self.branch, git_server, git_working_dir, callback=self.callback
+        )
 
     def __update_pull_plugin_artifact__(self, plugin_conf):
         """Pull the latest artifact to be sure we're up to date"""
         path = f"{self.iocroot}/jails/{self.jail}"
 
         shutil.rmtree(f"{path}/plugin", ignore_errors=True)
-        self.__clone_repo(plugin_conf['artifact'], f'{path}/plugin')
+        self._clone_repo(
+            self.branch, plugin_conf['artifact'],
+            f'{path}/plugin', callback=self.callback
+        )
 
         try:
             distutils.dir_util.copy_tree(
@@ -1403,12 +1422,12 @@ fingerprint: {fingerprint}
         fetch_args = {'release': release, 'eol': False}
         iocage_lib.iocage.IOCage(silent=self.silent).fetch(**fetch_args)
 
-    def __clone_repo(self, repo_url, destination, depth=None):
+    @staticmethod
+    def _clone_repo(ref, repo_url, destination, depth=None, callback=None):
         """
         This is to replicate the functionality of cloning/pulling a repo
         """
-        ref = self.branch
-
+        branch = ref
         try:
             # "Pull"
             repo = git.Repo(destination)
@@ -1452,12 +1471,13 @@ fingerprint: {fingerprint}
                     'level': 'EXCEPTION',
                     'message': f'Origin: {origin.url} does not exist!'
                 },
-                _callback=self.callback)
+                _callback=callback
+            )
 
         if f'origin/{ref}' not in repo.refs:
             ref = 'master'
             msgs = [
-                f'\nBranch {self.branch} does not exist at {repo_url}!',
+                f'\nBranch {branch} does not exist at {repo_url}!',
                 'Using "master" branch for plugin, this may not work '
                 'with your RELEASE'
             ]
@@ -1468,7 +1488,8 @@ fingerprint: {fingerprint}
                         'level': 'INFO',
                         'message': msg
                     },
-                    _callback=self.callback)
+                    _callback=callback
+                )
 
         # Time to make this reality
         repo.git.checkout(ref)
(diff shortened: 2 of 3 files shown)


william-gr (changes requested):
We need to improve this for speed.

Sequentially fetching each plugin json will be very slow, we should investigate either reuse the same connection to grab more files or perhaps fetch all at once using git (since we need the INDEX anayway, its probably better to just use git to get it all).

Same for packagesite, get a unique set and have a common method to fetch and extract it using a thread pool.

skarekrow:
For completeness sake, adding vlad to this (my review is the same as williams)

sonicaj:
@william-gr there can be a case when the pool hasn't been activated and we want to retrieve plugin versions. I suppose in this case we can check for activation of pool and if we don't find one, we retrieve the plugins files in a manner which reduces speed.

skarekrow:
We shouldn't have a use case like that @sonicaj. iocage requires a pool to do any operations, having a few exceptions just creates confusion on where to store artifacts.

sonicaj:
@skarekrow the idea is that we do all of this as soon as middlewared boots and cache the results instead of doing this the first time pool is activated and user actually is at the plugins screen waiting for the plugins to load

skarekrow:
@sonicaj We can do something special in the mw for that. This is for all of iocage, which isn't a good fit for a compromise like that

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +87,137 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
             # Backwards compat
             self.branch = 'master'
 
+    def clone_repo(self):
+        if self.server == "download.freebsd.org":
+            git_server = "https://github.com/freenas/iocage-ix-plugins.git"
+        else:
+            git_server = self.server
+
+        git_working_dir = f"{self.iocroot}/.plugin_index"
```
william-gr, line 93: Wont this require iocage to be activated to show a list of plugins?

sonicaj: That is the whole idea as @skarekrow suggested. I have separated the functions, once we have the plugin index and plugin json files data, that data can be passed on to the new function which retrieves versions.

iocage_lib/ioc_common.py
```diff
@@ -107,6 +108,78 @@ def logit(content, _callback=None, silent=False, exception=RuntimeError):
     _callback(content, exception)
 
 
+def clone_repo(ref, repo_url, destination, depth=None, callback=None):
```
skarekrow: I'm not so sure this belongs in common, after all, only the IOCPlugin module will ever be doing this.

skarekrow (changes requested):
We should store a json manifest that has a list of the plugins we had in the INDEX at last invocation. If they match, we should skip the fetching of package versions and use what is in that json manifest to save a lot of time and bandwidth. Otherwise this would happen each time the user requests the INDEX.

sonicaj:
@skarekrow we will be caching the results in mw, so that i believe is not needed here. If we ever decide to expose this in CLI, i believe it needs to be fresh. Also maintaining a list of plugins is not enough, versions can still be updated ;)

skarekrow:
Fair enough :P

iocage_lib/ioc_common.py
sonicaj: I can have this as a static method in the plugin interface, please advise what you think is better in this case. But it would seem weird calling `IOCPlugin._clone_repo`.

skarekrow: Well perhaps a different method name is appropriate, but it really should live in that module. It’s a private method for just it

sonicaj: Well this would be the exact invocation, `IOCPlugin._clone_repo(branch, repo_url, destination, depth, callback)` with params. Should i proceed ?

skarekrow: It's fine to me, @william-gr do you have a better suggestion for a method name?

william-gr: Looks ok to me

skarekrow (dismissed):

william-gr (commented):
I will wait to see how this will look in jail plugin before approve.

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +87,148 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+
+            return packagesite_url, package_site_data
+
+        plugin_packagesite_mapping = {}
+        package_sites = set([
+            url[:-1] if url[-1] == '/' else url for url in package_sites
+        ])
+
+        with concurrent.futures.ThreadPoolExecutor() as exc:
+            futures = exc.map(
```
william-gr: futures seems like a bad name, they are not futures, but an iterable of results?

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +87,148 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+            for future in futures:
+                plugin_packagesite_mapping[future[0]] = future[1]
+
+        return plugin_packagesite_mapping
+
+    @staticmethod
+    def fetch_plugin_versions_from_plugin_index(plugins_index):
+        plugin_packagesite_mapping = IOCPlugin.fetch_plugin_packagesites([
+            v['packagesite'] for v in plugins_index.values()
+            if v.get('packagesite')
```
william-gr: Is packagesite not required in plugin? If so what is the default we use and why its not in this list?

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +87,148 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+                    ) as f:
+                        packagesite_path = f.name
+                        for chunk in r.iter_content(chunk_size=8192):
+                            if chunk:
+                                f.write(chunk)
+            except Exception:
+                pass
+            else:
+                with tempfile.TemporaryDirectory() as td:
+                    tar = tarfile.open(packagesite_path)
```
william-gr: doesnt tarfile have context manager?

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +87,148 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+        version_dict = {}
+        for plugin in plugins_index:
+            plugin_dict = plugins_index[plugin]
+            packagesite = plugin_dict.get('packagesite')
+            primary_package = plugin_dict.get('primary_pkg') or plugin
+            if packagesite:
+                packagesite = packagesite[:-1] if packagesite[-1] == '/' \
+                    else packagesite
+                plugin_pkgs = plugin_packagesite_mapping[packagesite]
+                for pkg in plugin_pkgs:
```
william-gr: So we are iterating through every single package in each packagesite? Did I get this right?

That does not seem very good (O(n) for every plugin). We should come up with a way to make it O(log n) using a better structure.

This also likely means using two indexes because origin and package name, see the other discussion about origin.

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +87,148 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+                        os.path.join(td, 'packagesite.yaml'), 'rb'
+                    ) as f:
+                        for line in f.readlines():
+                            try:
+                                line_data = json.loads(line.decode())
+                            except UnicodeDecodeError:
+                                pass
+                            else:
+                                package_site_data[
+                                    line_data['origin']
```
william-gr: So I am confused about using origin.

How does this handle flavors?

A single origin can have multiple flavors and renders multiple different package names.

What if I want to use "unison-nox11" package from net/unison origin?

Perhaps we should start thinking about requiring pkg name and not origin in plugins? Since thats all it can use, packages and not ports.

iocage_lib/ioc_plugin.py
sonicaj: Ideally it is required. This is just better safe then sorry approach

iocage_lib/ioc_plugin.py
sonicaj: You're right. This will right now "apparently" work because we have origin's listed in primary package and not flavours. Now different flavours might have different versions which give unreliable results. I'll revisit the plugin INDEX and make sure we use pkg names instead of origin.

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +84,125 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+
+    @staticmethod
+    def fetch_plugin_packagesites(package_sites):
+        def download_parse_packagesite(packagesite_url):
+            package_site_data = {}
+            try:
+                response = requests.get(f'{packagesite_url}/All/')
+                response.raise_for_status()
+
+                for pkg in re.findall(r'<a.*>(.*).txz</a>', response.text):
```
william-gr: These are greedy, perhaps `r'<a.*?>(.*?).txz</a>'`, also wondering if you dont need re.S and re.M? i dont think is guaranteed to have them all on same line.

As even safer approach would be xml parsing (beautifulsoup).

william-gr (dismissed):
Much better ;) thanks

skarekrow (dismissed):

iocage_lib/ioc_plugin.py
sonicaj: @william-gr i tried xml parsing first, regex is 2x times faster. The way we have it right now, the flags aren't needed, but better safe then sorry, i'll add those ;)

william-gr: Just remembered  we probably dont want another dependency in iocage, so this should be fine :P

william-gr: Perhaps `r'\S+.txz'` ?

sonicaj: Yup, the regex flags are not needed in our case. Just a little more to the regex ;)

william-gr (dismissed):
Well, remember the html comes from the http server, so the output may differ from server to server (apache, nginx, lighttpd).

sonicaj:
The content between the tags should remain same ;)

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +84,125 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+                    },
+                    _callback=self.callback
+                )
+
+    @staticmethod
+    def fetch_plugin_packagesites(package_sites):
+        def download_parse_packagesite(packagesite_url):
+            package_site_data = {}
+            try:
+                response = requests.get(f'{packagesite_url}/All/')
```
themylogin: TImeout?

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +84,125 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+                    },
+                    _callback=self.callback
+                )
+
+    @staticmethod
+    def fetch_plugin_packagesites(package_sites):
+        def download_parse_packagesite(packagesite_url):
+            package_site_data = {}
+            try:
+                response = requests.get(f'{packagesite_url}/All/', timeout=120)
```
themylogin: 120?

sonicaj: Should i decrease it to 60 perhaps ?

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +84,125 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+
+    @staticmethod
+    def fetch_plugin_packagesites(package_sites):
+        def download_parse_packagesite(packagesite_url):
+            package_site_data = {}
+            try:
+                response = requests.get(f'{packagesite_url}/All/', timeout=20)
+                response.raise_for_status()
+
+                for pkg in re.findall(r'<a.*>\s*(\S+).txz</a>', response.text):
```
themylogin, line 118: Please use `html.parser` module instead

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +84,125 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+
+                    pkg, version = pkg.rsplit('-', 1)
+                    epoch_split = version.rsplit(',', 1)
+                    epoch = epoch_split[1] if len(epoch_split) == 2 else '0'
+                    revision_split = epoch_split[0].rsplit('_', 1)
+                    revision = \
+                        revision_split[1] if len(revision_split) == 2 else '0'
+                    package_site_data[pkg] = {
```
themylogin: How about moving this (parsing `pkg` into dict) to to separate function and creating some unit tests?

iocage_lib/ioc_plugin.py
```diff
@@ -84,6 +84,125 @@ def __init__(self, release=None, jail=None, plugin=None, branch=None,
...
+                        'epoch': epoch,
+                    }
+            except Exception:
+                pass
+
+            return packagesite_url, package_site_data
+
+        plugin_packagesite_mapping = {}
+        package_sites = set([
+            url[:-1] if url[-1] == '/' else url for url in package_sites
```
themylogin: `rstrip("/")`?

themylogin (changes requested):

iocage_lib/ioc_plugin.py
sonicaj: I did, it's way slower

themylogin: I've got it wrong. Let's use `etree.HTMLParser` from `lxml`

william-gr: We cant add yet another dependency to iocage only for this

themylogin: If lxml is not a dependency of iocage then fine, let's leave it like this @sonicaj

themylogin (dismissed):
Thank you!

skarekrow (approved):

PR #954, opened 2019-06-17
Allow Ipv6 only jails on IPv6 only hosts

joachimmathes:
### Bug description ###

  `Version 1.2 RC (40327cdb6afbbb725e452d2709887e1a221ffa00)`
  `iocage start <jail>`
  ```
  No default gateway interface found
  ```
  ```
  * Starting vprot01
    + Started OK
    + Using devfs_ruleset: 14
    + Configuring VNET OK
    + Using IP options: vnet
    + Starting services OK
    + Executing poststart OK
    + Setting RCTL props
  ```
  a stable release.

---

### Description of PR ###

We have setups where IPv6 only jails reside on IPv6 only hosts. Thus,
we set
```
defaultrouter='none'
defaultrouter6='fe80::ff%epair0b'
```
This is where
https://github.com/iocage/iocage/blob/40327cdb6afbbb725e452d2709887e1a221ffa00/iocage_lib/ioc_start.py#L1315
raises an exception since IPv6 routing is not respected by
https://github.com/iocage/iocage/blob/40327cdb6afbbb725e452d2709887e1a221ffa00/iocage_lib/ioc_start.py#L216
.

For now, as a workaround, we can set
```
defaultrouter=''
```
. But I don't think its meaning becomes very clear, especially in contrast to `none`.

So, this PR
1. introduces the default value `auto` for `defaultrouter` and
   `defaultrouter6`. (I know, this is a breaking change.)
2. uses system command `netstat` instead of python module `netifaces`,
   since `netstat` is more reliable. `netifaces` does not return IPv6
   default routers in our environment.

What do you think?

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -700,7 +700,7 @@ class IOCConfiguration(IOCZFS):
     @staticmethod
     def get_version():
         """Sets the iocage configuration version."""
-        version = '23'
+        version = '24'
 
         return version
 
@@ -947,11 +947,6 @@ class IOCConfiguration(IOCZFS):
         if current_conf_version == iocage_conf_version:
             return conf, False
 
-        # New style thin configuration jails won't have this. Only their
-        # defaults will
-        if current_conf_version is None and thickconfig != 'THICK':
-            return conf, False
-
         if os.geteuid() != 0:
             iocage_lib.ioc_common.logit(
                 {
@@ -964,10 +959,13 @@ class IOCConfiguration(IOCZFS):
                 silent=self.silent,
                 exception=ioc_exceptions.CommandNeedsRoot)
 
-        if not default:
-            jail_conf = self.check_jail_config(conf)
-
         conf['CONFIG_VERSION'] = iocage_conf_version
+
+        # New style thin configuration jails won't have this. Only their
+        # defaults will
+        if thickconfig != 'THICK' and not default:
+            jail_conf = self.check_jail_config(conf)
+            return jail_conf, False
 
         # Version 2 keys
         if not conf.get('sysvmsg'):
@@ -1070,6 +1068,12 @@ class IOCConfiguration(IOCZFS):
             conf['nat_backend'] = 'ipfw'
         if not conf.get('nat_forwards'):
             conf['nat_forwards'] = 'none'
+
+        # Version 24 keys
+        # Migrate defaultrouter and defaultrouter6 default 'none' to 'auto'
+        for option in ('defaultrouter', 'defaultrouter6'):
+            if conf.get(option) == 'none':
+                conf[option] = 'auto'
 
         if not default:
             conf.update(jail_conf)
@@ -1191,6 +1195,12 @@ class IOCConfiguration(IOCZFS):
             # New jail creation
             pass
 
+        # Version 24 keys
+        # Migrate defaultrouter and defaultrouter6 default 'none' to 'auto'
+        for option in ('defaultrouter', 'defaultrouter6'):
+            if conf.get(option) == 'none':
+                del conf[option]
+
         try:
             if not renamed:
                 self.json_write(conf)
@@ -1265,8 +1275,8 @@ class IOCConfiguration(IOCZFS):
             'ip6_addr': 'none',
             'ip6_saddrsel': '1',
             'ip6': 'new',
-            'defaultrouter': 'none',
-            'defaultrouter6': 'none',
+            'defaultrouter': 'auto',
+            'defaultrouter6': 'auto',
             'resolver': '/etc/resolv.conf',
             'mac_prefix': self.mac_prefix,
             'vnet0_mac': 'none',
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -62,8 +62,8 @@ class IOCStart(object):
         self.unit_test = unit_test
         self.ip4_addr = 'none'
         self.ip6_addr = 'none'
-        self.defaultrouter = 'none'
-        self.defaultrouter6 = 'none'
+        self.defaultrouter = 'auto'
+        self.defaultrouter6 = 'auto'
         self.log = logging.getLogger('iocage')
 
         if not self.unit_test:
@@ -174,6 +174,7 @@ class IOCStart(object):
         localhost_ip = self.conf['localhost_ip']
         self.defaultrouter = self.conf['defaultrouter']
         self.defaultrouter6 = self.conf['defaultrouter6']
+        self.host_gateways = iocage_lib.ioc_common.get_host_gateways()
 
         fstab_list = []
         with open(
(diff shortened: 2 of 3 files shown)


iocage_lib/ioc_start.py
```diff
@@ -1312,19 +1311,45 @@ def __check_rtsold__(self):
...
+    def get_default_ipv6_gateway(self):
+        gateway = ''
+        if netifaces.AF_INET6 in self.host_gateways['default']:
+            gateway = re.sub(
+                '%.*$', '%epair0b',
+                self.host_gateways['default'][netifaces.AF_INET6][0])
+        iocage_lib.ioc_common.logit({
+            'level': 'WARNING',
+            'message': f'{self.uuid}: vnet requires IPv6 defaultrouter,'
+            f' using {gateway}'
```
skarekrow: This may look a little odd if gateway is actually empty, we should always show them something informative.

iocage_lib/ioc_start.py
```diff
@@ -1312,19 +1311,45 @@ def __check_rtsold__(self):
...
+                silent=self.silent)
+
+    def get_default_ipv4_gateway(self):
+        gateway = ''
+        if netifaces.AF_INET in self.host_gateways['default']:
+            gateway = self.host_gateways['default'][netifaces.AF_INET][0]
+        iocage_lib.ioc_common.logit({
+            'level': 'WARNING',
+            'message': f'{self.uuid}: vnet requires IPv4 defaultrouter,'
+            f' using {gateway}'
```
skarekrow: Same as below

iocage_lib/ioc_start.py
```diff
@@ -213,17 +214,15 @@ def __start_jail__(self):
             }, _callback=self.callback,
                 silent=self.silent)
 
-        if self.conf['vnet'] and self.defaultrouter == 'none':
-            self.log.debug('Grabbing default route')
-            self.defaultrouter = self.get_default_gateway()[0]
-            self.log.debug(f'Default Gateway: {self.defaultrouter}')
+        if self.conf['vnet'] and self.defaultrouter == 'auto':
+            self.log.debug('Grabbing IPv4 default route')
+            self.defaultrouter = self.get_default_ipv4_gateway()
```
skarekrow, line 219: We should collapse `ipv4_gateway` and `ipv6_gateway` and instead supply a string to check against. Since they are very similar code blocks, with 6 requiring just a couple extra things.

iocage_lib/ioc_start.py
```diff
@@ -213,17 +214,15 @@ def __start_jail__(self):
             }, _callback=self.callback,
                 silent=self.silent)
 
-        if self.conf['vnet'] and self.defaultrouter == 'none':
-            self.log.debug('Grabbing default route')
-            self.defaultrouter = self.get_default_gateway()[0]
-            self.log.debug(f'Default Gateway: {self.defaultrouter}')
+        if self.conf['vnet'] and self.defaultrouter == 'auto':
```
skarekrow, line 217: While auto makes more sense, we need a migration in place for those who are already using none. DHCP and NAT users specifically will run into issues.

skarekrow (changes requested):
Thanks for the PR @joachimmathes! @sonicaj thoughts?

iocage_lib/ioc_start.py
joachimmathes: You are right. This message only made sense in the `if self.conf['vnet'] and self.defaultrouter == 'none'` context:
https://github.com/iocage/iocage/blob/40327cdb6afbbb725e452d2709887e1a221ffa00/iocage_lib/ioc_start.py#L216

iocage_lib/ioc_start.py
joachimmathes: Yup

iocage_lib/ioc_start.py
joachimmathes: Do you mean merging
```
get_default_ipv4_gateway(self)
```
and
```
get_default_ipv6_gateway(self)
```
into one method with a signature like
```
get_default_ipv6_gateway(self, address_family=netifaces.AF_INET)
```
?

This was my first approach, but with respect to clean code patterns, I
decided to separate them:
1. Let the method name speak for itself instead of its probably
   obscure parameters.
2. Keep methods as small as possible and separate their concerns, even
   if they are similar.

But if you prefer a single method, I will implement it. In this case
it's a subtleness, anyway.

iocage_lib/ioc_start.py
joachimmathes: Yeah, this is definitely a breaking change. I will provide a migration.

joachimmathes:
It was not so easy to get a grip on the migration code. It seems that https://github.com/iocage/iocage/blob/219b1858e905a1e4984736edfc82939911eca4da/iocage_lib/ioc_json.py#L952 prohibited any jail config updates. I tried to fix it and as far as our jails are concerned it works.

iocage_lib/ioc_json.py
```diff
@@ -1191,6 +1195,12 @@ def check_jail_config(self, conf):
             # New jail creation
             pass
 
+        # Version 24 keys
```
skarekrow, line 1198: These aren't needed, the check_jail_config is simply for legacy jails now.

iocage_lib/ioc_json.py
```diff
@@ -964,11 +959,14 @@ def check_config(self, conf, default=False):
...
                 exception=ioc_exceptions.CommandNeedsRoot)
 
-        if not default:
-            jail_conf = self.check_jail_config(conf)
-
         conf['CONFIG_VERSION'] = iocage_conf_version
 
+        # New style thin configuration jails won't have this. Only their
+        # defaults will
+        if thickconfig != 'THICK' and not default:
```
skarekrow, line 966: This will result in incorrect behavior, we want to return early as the code had it as if it's not a THICK configuration type, the default value is actually the one that gets updated, not each thin jail. Thin jails by design only have the changes the user requested and a few other essentials. So what you were likely seeing is the default property changing, and not the jails. So please revert this change.

This code block is used for THICK configuration types, and default configurations. Which all thin jails inherit from.

iocage_lib/ioc_start.py
skarekrow: Yeah, since I like the principle of DRY, I try to keep similar functions together. So I suggest a method called `get_default_gateway(self, address_family=netifaces.AF_INET)` and have it default like you did to ipv4, if the address family is 6, then do the 6 stuff as well.

skarekrow (changes requested):

joachimmathes:
Hi Brandon,

I'm very sorry, but I have to close this pull request. 

We operate hundreds of jails with diverse and complex network setups in production. Unfortunately, keeping track of further changes (and regressions) in iocage with respect to our environment is not possible at the moment. We simply do not have the capacities. Furthermore, since we use only few iocage features, it's actually feature complete for us. We might even take things further and replace iocage by basic FreeBSD jail tools or very stripped down glue code. So, for now my team has decided to close this PR and use an iocage fork instead.

Best
Joachim

sonicaj:
Hello @joachimmathes, i am sad to see you go away. For this however, we are more then ready to help you finish this PR ( even take over entirely if you want ). I'll still be creating a ticket for this in our tracker, but without a ideal testing env, i'll have to punt this for a while. If you however feel like helping us here, we can have this fixed earlier with your testing env.
Thank you for bringing this to our attention.
Cheers,
Waqar

joachimmathes:
Hello @sonicaj and @skarekrow ,

I'm going to contact you next weekend to solve this PR.

See you then :smiley:
Jo

PR #955, opened 2019-06-19
NAS-101922 / 11.3 / Migrate iocage mountpoints

sonicaj:
This commit adds a check in check interface of iocage which would perform migration of iocage dataset mountpoint if it is not mounted under the activated pool. This would default to false for FreeBSD users and for FN/TN users, we would enforce this.


william-gr (approved):
Looks reasonable, but wait for brandon's approval

skarekrow (approved):

PR #957, opened 2019-06-19
NAS-102168 / 11.3 / Normalize plugin fetch output

sonicaj:
This commit normalizes plugin fetch output to use a dictionary to return plugin data instead of list which is ambiguous.

### DIFF
--- a/iocage_lib/ioc_plugin.py
+++ b/iocage_lib/ioc_plugin.py
@@ -797,41 +797,47 @@ fingerprint: {fingerprint}
         if index_only:
             return plugins
 
-        _plugins = self.__fetch_sort_plugin__(plugins, official=official)
+        plugins_ordered_dict = collections.OrderedDict(
+            sorted({
+                k: {'name': v['name'], 'description': v['description']}
+                for k, v in plugins.items()
+                if not (official and not v.get('official', False))
+            }.items())
+        )
 
         if self.plugin is None and not _list:
-            for p in _plugins:
+            for i, p in enumerate(plugins_ordered_dict.items()):
+                k, v = p
                 iocage_lib.ioc_common.logit(
                     {
-                        "level": "INFO",
-                        "message": f"[{_plugins.index(p)}] {p}"
+                        'level': 'INFO',
+                        'message':
+                            f'[{i}] {v["name"]} - {v["description"]} ({k})'
                     },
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_plugin.py
```diff
@@ -846,6 +852,17 @@ def fetch_plugin_index(
                 if icon:
                     list_header += ["ICON"]
 
+                plugin_list = [
```
themylogin, line 855: I am a little out of context, so sorry if I am saying nonsense, but you seem to be removing cells with empty values from table rows, shifting rightmost cells to the left, breaking table structure (cols won't match neither with headers nor with cols in other rows), is this what you are doing?

sonicaj: @themylogin nope. `name`, `description`, `pkg` are always present. `icon` is controlled by variable, so if `icon` is false, the dictionary won't have it, and it won't be included in the list for the table. Table structure would remain intact that way with/without icon supplied.

iocage_lib/ioc_plugin.py
```diff
@@ -887,58 +905,23 @@ def __fetch_validate_plugin__(self, plugin, plugins):
...
             try:
-                plugin = [
-                    i for i, p in enumerate(plugins)
-
-                    if plugin.capitalize() in p or plugin in p
-                ]
-                try:
-                    plugin = plugins[int(plugin[0])]
-                except IndexError:
+                plugins[plugin]
```
william-gr: Why force an exception if you could just do `if plugin not in plugins:` ?

william-gr (dismissed):

iocage_lib/ioc_plugin.py
sonicaj: Updated @william-gr

william-gr (dismissed):

william-gr (dismissed):

iocage_lib/ioc_plugin.py
themylogin: Can it be more clear then? What protects us from empty `description`?

sonicaj: @themylogin updated

themylogin (approved):
LGTM, thank you!

skarekrow (approved):

PR #958, opened 2019-06-20
NAS-102205 / 11.3 / Prop for plugin jails

sonicaj:
This PR introduces a property which will keep track of which plugin a jail represents even after it has been renamed.

### DIFF
--- a/iocage_lib/ioc_plugin.py
+++ b/iocage_lib/ioc_plugin.py
@@ -53,6 +53,8 @@ class IOCPlugin(object):
     includes creation, updating and upgrading.
     """
 
+    PLUGIN_VERSION = '2'
+
     def __init__(self, release=None, jail=None, plugin=None, branch=None,
                  keep_jail_on_failure=False, callback=None, silent=False,
                  **kwargs):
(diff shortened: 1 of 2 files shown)


themylogin (dismissed):

iocage_lib/ioc_plugin.py
```diff
@@ -53,6 +53,8 @@ class IOCPlugin(object):
     includes creation, updating and upgrading.
     """
 
+    VERSION = 'pluginv2'
```
skarekrow: Perhaps just 2 here and rename the variable to PLUGIN_VERSION

iocage_lib/ioc_plugin.py
```diff
@@ -435,13 +437,22 @@ def __fetch_plugin_props__(self, conf, props, num):
...
+        ]
 
         create_props = create_props + [
             f"{k}={v}" for k, v in (p.split("=") for p in props)
         ]
 
+        # These properties are not user configurable
+
+        for prop in (
+            f'type={self.VERSION}',
```
skarekrow: Then put pluginv{self.PLUGIN_VERSION} here

skarekrow (changes requested):

skarekrow (approved):

PR #959, opened 2019-06-20
NAS-101743 / 11.3 / Use global variable INTERACTIVE instead of is_tty

skarekrow:
This allows us to retain the original exceptions in iocage

NAS-101743


iocage_lib/ioc_common.py
```diff
@@ -69,25 +69,27 @@ def callback(_log, callback_exception):
         log.log(25, message)
     elif level == 'EXCEPTION':
         try:
-            if not is_tty():
+            interactive = INTERACTIVE
+        except NameError:
```
themylogin: Why not define `INTERACTIVE = False` at the top of this file?

themylogin (changes requested):

iocage_lib/ioc_common.py
themylogin: Now this NameError is not possible

themylogin (dismissed):
LGTM, thank you!

iocage_lib/iocage.py
```diff
@@ -867,7 +867,7 @@ def exec(
         status, jid = self.list("jid", uuid=uuid)
 
         if not status and not start_jail:
-            if not ioc_common.is_tty():
+            if not INTERACTIVE:
```
sonicaj: @skarekrow i think it would be better if we are explicit about this, it would work because when `ioc_common` is imported, the variable is set but it would still show up as an error on IDE's as it is unable to find a reference. I would recommend using `ioc_common.INTERACTIVE`

iocage_lib/ioc_plugin.py
```diff
@@ -533,9 +533,10 @@ def __fetch_plugin_install_packages__(self, jaildir, conf, pkg_repos,
                 0,
                 pkglist=["ca_root_nss"],
                 silent=True,
-                callback=self.callback).create_install_packages(
+                callback=self.callback
+            ).create_install_packages(
```
sonicaj, line 537: Please could you fix the indentation below as well ;)

sonicaj (approved):
Thank you

PR #960, opened 2019-06-21
NAS-101249 / 11.3 / Bug fix for spaces in path names for fstab

sonicaj:
This commit fixes an issue where if spaces were defined in src/destination paths, we could not cleanly interpret where to split.


skarekrow (approved):

skarekrow:
We should backport this to 11.2 as well, @william-gr thoughts? They've been in lockstep fstab wise.

Issue #952, opened 2019-06-17
Ruleset exists but iocage does not find it


dlangille:
This worked before upgrading `iocage` today and upgrading the jail to `12.0-RELEASE-p5`

Make sure to follow and check these boxes before submitting an issue! Thank you.

```
$ iocage --version
Version	1.2 RC

```
The above is head from 2019-06-16:

```
$ pkg info -x iocage
py36-iocage-devel-1.0.0.20190616,1
```

Today I upgraded `iocage` from 1.0.0.20190519_1,1

When the jail is started, the rule set is not invoked/found:

```
[dan@knew:~] $ date
Mon Jun 17 01:17:55 UTC 2019
[dan@knew:~] $ sudo iocage start toiler
* Ruleset 7 does not exist, using defaults
* Starting toiler
  + Started OK
  + Using devfs_ruleset: 7
  + Using IP options: ip4.addr=10.55.0.13,127.1.0.53 ip6.addr=2001:470:8abf:[redacted] ip4.saddrsel=1 ip4=new ip6.saddrsel=1 ip6=new
  + Starting services OK
  + Executing poststart OK


[dan@knew:~] $ tail /var/log/iocage.log 
2019/06/17 01:17:01 (INFO)   + Removing devfs_ruleset: 7 OK
2019/06/17 01:17:01 (INFO)   + Removing jail process OK
2019/06/17 01:17:01 (INFO)   + Executing poststop OK
2019/06/17 01:18:05 (INFO) * Ruleset 7 does not exist, using defaults
2019/06/17 01:18:05 (INFO) * Starting toiler
2019/06/17 01:18:05 (INFO)   + Started OK
2019/06/17 01:18:05 (INFO)   + Using devfs_ruleset: 7
2019/06/17 01:18:05 (INFO)   + Using IP options: ip4.addr=10.55.0.13,127.1.0.53 ip6.addr=2001:470:8abf:[redacted] ip4.saddrsel=1 ip4=new ip6.saddrsel=1 ip6=new
2019/06/17 01:18:08 (INFO)   + Starting services OK
2019/06/17 01:18:08 (INFO)   + Executing poststart OK
[dan@knew:~] $ 

```

NOTE the 'does not exist' in the log extra above.

Within `/etc/devfs.rules` we have

```
# for dhcpd in jail toiler
#
[devfsrules_jail_bpf=7]
add include $devfsrules_hide_all
add include $devfsrules_unhide_basic
add include $devfsrules_unhide_login
[...]

PR #961, opened 2019-06-23
Allow IPv6 only jails on IPv6 only hosts

joachimmathes:
See https://github.com/iocage/iocage/pull/954 for PR history.


joachimmathes:
### `get_default_gateway()` and a few words on code in general ###

I joined `get_default_ipv4_gateway()` and `get_default_ipv6_gateway()` into
`get_default_gateway()`.

Again, I am not happy with this solution. It might look like a simplification
and prevention of DRY. But in my opinion it's a code smell. As I wrote:
* Let the method name speak for itself instead of its probably
  obscure parameters.
* Keep methods as small as possible and separate their concerns, even
  if they are similar.

IPv4 and IPv6 demand different treatment, in this case suffix
replacement. That's why I don't consider this a DRY violation. We are lucky
that it works for both IPv6 and IPv4. But, for the method user, it is not
clear at first sight that `get_default_gateway` handles two types of
gateways. And on the other side, the method implementer has to understand
that the suffix substituion is valid for both address families, but is
actually not needed for IPv4.

Sorry, for this elaboration. I know, in this case it might be a subtlety. But
to be honest, I have a really hard time to understand what the iocage code
does:
* Class `IOCStart` is more than 1500 lines long.
* `__start_jail__()` has about 785 lines
* containing complex `if, for, try, for, try, if` nestings.

IMHO the two mentioned patterns would help to improve the code a lot.

iocage_lib/ioc_json.py
```diff
@@ -964,11 +959,14 @@ def check_config(self, conf, default=False):
...
                 exception=ioc_exceptions.CommandNeedsRoot)
 
-        if not default:
-            jail_conf = self.check_jail_config(conf)
-
         conf['CONFIG_VERSION'] = iocage_conf_version
 
+        # New style thin configuration jails won't have this. Only their
+        # defaults will
+        if thickconfig != 'THICK' and not default:
```
joachimmathes: @skarekrow @sonicaj I haven't removed my changes, yet, here. And maybe this is the wrong place in the code. But I think, when we change the behavior of `none`, we have to consider jails where `defaultrouter=none` has already been set by an iocage user explicitly. In these cases `defaultrouter=auto` of the "thick" default configuration would be overwritten by `defaultrouter=none` of the "thin" jail specific configuration.

sonicaj:
@joachimmathes i have a query, when you have some time please let me know what you think.
You mention that `netifaces` does not return ipv6 default routers. I am afraid i can't find any reference for this and i can't test it on my own network so i am not sure. Can you please let me know how you verified this ? ( i am not against using netstat but i would like to be sure we have good reason to )

iocage_lib/ioc_json.py
sonicaj: @joachimmathes you introduced a regression with this change for base jails. Please see the tests ;)

iocage_lib/ioc_json.py
```diff
@@ -1071,6 +1069,12 @@ def check_config(self, conf, default=False):
...
             conf['nat_forwards'] = 'none'
 
+        # Version 24 keys
+        # Migrate defaultrouter and defaultrouter6 default 'none' to 'auto'
+        for option in ('defaultrouter', 'defaultrouter6'):
+            if conf.get(option) == 'none':
+                conf[option] = 'auto'
+
         if not default:
             conf.update(jail_conf)
```
sonicaj, line 1097: `jail_conf` does not exist because it got removed earlier in the same file

iocage_lib/ioc_json.py
```diff
@@ -964,11 +959,14 @@ def check_config(self, conf, default=False):
                 silent=self.silent,
                 exception=ioc_exceptions.CommandNeedsRoot)
 
-        if not default:
-            jail_conf = self.check_jail_config(conf)
```
sonicaj: This is where it's removed

iocage_lib/ioc_json.py
sonicaj: Also we understand your perspective on this and it is absolutely right. We have opened up a new ticket for that which will improve how we keep track of props as to when it's changed if at all. For now, let's please stick to what @skarekrow suggested in the other PR because it is going to cause inconsistent behaviour with thin type configs.

iocage_lib/ioc_json.py
```diff
@@ -1071,6 +1069,12 @@ def check_config(self, conf, default=False):
         if not conf.get('nat_forwards'):
             conf['nat_forwards'] = 'none'
 
+        # Version 24 keys
```
sonicaj: This would be version 25 now and also you forgot to bump the config version

iocage_lib/ioc_common.py
```diff
@@ -1032,3 +1032,32 @@ def parse_package_name(pkg):
...
+
+    output = su.run(['netstat', '-r', '-n', '--libxo', 'json'],
+                    stdout=su.PIPE)
+    route_families = (json.loads(output.stdout)
+                      ['statistics']
+                      ['route-information']
+                      ['route-table']
+                      ['rt-family'])
+
+    gateways = {'default': {}}
```
sonicaj: We are only interested in retrieving `default`, do you think perhaps it would be better to map `netifaces.AF_INET\6` directly instead of accessing `default` key always before ?

iocage_lib/ioc_common.py
```diff
@@ -1032,3 +1032,32 @@ def parse_package_name(pkg):
...
+    for af in ('Internet', 'Internet6'):
+        route_entries = list(filter(
+            lambda x: x['address-family'] == af, route_families)
+        )[0]['rt-entry']
+        default_route = list(filter(
+            lambda x: x['destination'] == 'default', route_entries)
+        )
+        if default_route and 'gateway' in default_route[0]:
+            gateways['default'][netifaces_af_mapping[af]] = (
+                default_route[0]['gateway'], default_route[0]['interface-name']
```
sonicaj: While we are at it, `keys` are always better then using `indexes` to refer items as the former is clearer, i would recommend we use a dict here

iocage_lib/ioc_start.py
```diff
@@ -1312,19 +1311,27 @@ def __check_rtsold__(self):
...
                 },
                 _callback=self.callback,
-                silent=self.silent
-            )
+                silent=self.silent)
+
+    def get_default_gateway(self, address_family=netifaces.AF_INET):
+        if address_family in self.host_gateways['default']:
+            return re.sub(
+                '%.*$', '%epair0b',
```
sonicaj: I am sorry, i do not follow this change, could you please let me know what purpose this serves ?

iocage_lib/ioc_start.py
```diff
@@ -1312,19 +1311,27 @@ def __check_rtsold__(self):
...
+                self.host_gateways['default'][address_family][0])
+        else:
+            return ''
```
sonicaj: This should return `none` i think which will indicate nothing can be found for this and hence we don't perform anything for `none` in our starting setup

sonicaj: This would also make sure that we don't set empty string on start up

sonicaj (commented):
I can understand @joachimmathes that the class is indeed quite long. But it's been quite some time when this was written if you go back in history ;) We have plans to restructure it but all in due time ;)
Just a few notes.

sonicaj:
@joachimmathes if you think we should take over, please feel free to let us know and we would be happy to help. Once being done with the changes, you can test the PR for us and we can then merge if everything seems good. Looking forward to hearing from you.
Cheers!

joachimmathes:
@sonicaj I will have a closer look at the `netifaces` module (and your PR comments) tomorrow, when I have my production environment around. :smiley:

joachimmathes:
> @joachimmathes i have a query, when you have some time please let me know what you think.
> You mention that `netifaces` does not return ipv6 default routers. I am afraid i can't find any reference for this and i can't test it on my own network so i am not sure. Can you please let me know how you verified this ? ( i am not against using netstat but i would like to be sure we have good reason to )

@sonicaj it's weirder than I though. This is the `netifaces` package we use on all hosts:
```
py36-netifaces-0.10.9
Name           : py36-netifaces
Version        : 0.10.9
Installed on   : Tue May  7 14:06:21 2019 CEST
Origin         : net/py-netifaces
Architecture   : FreeBSD:11:amd64
Prefix         : /usr/local
Categories     : python net
Licenses       : MIT
Maintainer     : bofh@FreeBSD.org
WWW            : https://pypi.org/project/netifaces/
Comment        : Getting network addresses from Python 3
Shared Libs required:
        libpython3.6m.so.1.0
Annotations    :
        FreeBSD_version: 1102000
        flavor         : py36
        repo_type      : binary
        repository     : mariadb100-punkt.de
Flat size      : 26.5KiB
Description    :
Getting network addresses from Python

It's been irritating me for ages that it isn't possible to
straightforwardly get the network address(es) of the machine
you're running on from a Python program.

WWW: https://pypi.org/project/netifaces/
```

On one host we get:
```
$ python3
Python 3.6.8 (default, Apr  2 2019, 01:15:55)
[GCC 4.2.1 Compatible FreeBSD Clang 6.0.0 (tags/RELEASE_600/final 326565)] on freebsd11
Type "help", "copyright", "credits" or "license" for more information.
>>> import netifaces
>>> print(netifaces.gateways())
{'default': {}}
```

On another one we get:
```
$ python3
Python 3.6.8 (default, May 17 2019, 23:49:25)
[GCC 4.2.1 Compatible FreeBSD Clang 6.0.0 (tags/RELEASE_600/final 326565)] on freebsd11
Type "help", "copyright", "credits" or "license" for more information.
>>> import netifaces
>>> print(netifaces.gateways())
{'default': {248: ('ff:ff:1c:1c:00:00:00:00:00:00:fe:80:00:00:00:00:00:00:00:00:00:00:00', 'inet0')}, 248: [('ff:ff:1c:1c:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00', 'lo0', True), ('ff:ff:1c:1c:00:00:00:00:00:00:fe:80:00:00:00:00:00:00:00:00:00:00:00', 'inet0', True)]}
```

This is complete nonsense, since both hosts have a valid IPv6 default route. They are link local addresses. Maybe that's the reason. Don't know. :man_shrugging:

If you decide to keep my `netstat` implementation, I going to implement your review comments.

sonicaj:
@joachimmathes this is certainly weird. I have been reading up on link local addressing and ipv6 and it seems they should have a default gateway always with link local defaulting to `fe80::1`.
Can you please share `netstat` outputs on the same hosts you ran `netifaces` commands ? ( marking which doesn't give a gateway with `netifaces` )
I'll try to have someone else peek at this too to make sure we don't miss something important and then we can proceed with the PR.
Looking forward to hearing from you.
Cheers!

joachimmathes:
@sonicaj here we go:

### Example 1: ###

```
$ python3
Python 3.6.8 (default, Apr  2 2019, 01:15:55)
[GCC 4.2.1 Compatible FreeBSD Clang 6.0.0 (tags/RELEASE_600/final 326565)] on freebsd11
Type "help", "copyright", "credits" or "license" for more information.
>>> import netifaces
>>> print(netifaces.gateways())
{'default': {}}
```
belongs to
```
$ netstat -rn
Routing tables

Internet:
Destination        Gateway            Flags     Netif Expire
10.5.0.0/16        link#5             U         mgmt0
10.5.103.28        link#5             UHS         lo0
127.0.0.1          link#3             UH          lo0

Internet6:
Destination                       Gateway                       Flags     Netif Expire
::/96                             ::1                           UGRS        lo0
default                           fe80::12%inet0                UGS       inet0
::1                               link#3                        UH          lo0
::ffff:0.0.0.0/96                 ::1                           UGRS        lo0
64:ff9b::/96                      2a00:b580:8000:12:c190:b6be:3794:6666 UGS    inet0
2a00:b580:8000:12::/64            link#4                        U         inet0
2a00:b580:8000:12:295c:642f:c1d2:7943 link#4                    UHS         lo0
fe80::/10                         ::1                           UGRS        lo0
fe80::%lo0/64                     link#3                        U           lo0
fe80::1%lo0                       link#3                        UHS         lo0
fe80::%inet0/64                   link#4                        U         inet0
fe80::ae1f:6bff:fe6d:80a7%inet0   link#4                        UHS         lo0
fe80::%vnet0.7/64                 link#8                        U       vnet0.7
fe80::1013:28ff:fe88:484b%vnet0.7 link#8                        UHS         lo0
ff02::/16                         ::1                           UGRS        lo0
```

### Example 2 ###

```
$ python3
Python 3.6.8 (default, May 17 2019, 23:49:25)
[GCC 4.2.1 Compatible FreeBSD Clang 6.0.0 (tags/RELEASE_600/final 326565)] on freebsd11
Type "help", "copyright", "credits" or "license" for more information.
>>> import netifaces
>>> print(netifaces.gateways())
{'default': {248: ('ff:ff:1c:1c:00:00:00:00:00:00:fe:80:00:00:00:00:00:00:00:00:00:00:00', 'inet0')}, 248: [('ff:ff:1c:1c:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00', 'lo0', True), ('ff:ff:1c:1c:00:00:00:00:00:00:fe:80:00:00:00:00:00:00:00:00:00:00:00', 'inet0', True)]}
```
belongs to
```
$ netstat -rn
Routing tables

Internet:
Destination        Gateway            Flags     Netif Expire
10.5.0.0/16        link#9             U         mgmt0
10.5.105.7         link#9             UHS         lo0
127.0.0.1          lo0                UHS         lo0

Internet6:
Destination                       Gateway                       Flags     Netif Expire
::/96                             ::1                           UGRS        lo0
default                           fe80::11%inet0                UGS       inet0
::1                               lo0                           UHS         lo0
::ffff:0.0.0.0/96                 ::1                           UGRS        lo0
64:ff9b::/96                      2a00:b580:8000:11:7de1:623f:2a2f:ad51 UGS    inet0
2a00:b580:8000:11::/64            link#8                        U         inet0
2a00:b580:8000:11:44e8:ab80:816:7869 link#8                     UHS         lo0
fe80::/10                         ::1                           UGRS        lo0
fe80::%lo0/64                     link#7                        U           lo0
fe80::1%lo0                       link#7                        UHS         lo0
fe80::%inet0/64                   link#8                        U         inet0
fe80::92e2:baff:feed:87bc%inet0   link#8                        UHS         lo0
fe80::%mgmt0/64                   link#9                        U         mgmt0
fe80::92e2:baff:feed:87bd%mgmt0   link#9                        UHS         lo0
fe80::%vnet0.253/64               link#35                       U      vnet0.25
fe80::1015:7ff:febc:1c50%vnet0.253 link#35                      UHS         lo0
ff02::/16                         ::1                           UGRS        lo0
```

sonicaj:
@joachimmathes please proceed with the `netstat` implementation. I haven't been able to find anything which would point us in a direction with `netifaces`. I assume it might just be broken in this case. Please do have a look at other comments i added before. Also config version would be 26 now i believe because of more changes recently merged to ioc_json.
Cheers!

iocage_lib/ioc_start.py
joachimmathes: Do you mean the `re.sub`?

As you can see in the `netstat` output down there `defaultrouter6 ` is a link local address with a `%inet0` suffix. This has to be replaced by the corresponding `%epairXb` suffix. For now `X = 0` is a good guess. In another environment it could be `X > 0`, of course.

iocage_lib/ioc_common.py
joachimmathes: Yeah, you are right. I wanted to keep up the `netifaces` API. But with only one default value being returned this is very artificial.

iocage_lib/ioc_common.py
joachimmathes: See above. I will change that.

sonicaj: Yup, something like the following would be nice:
```
{"ipv4": {"gateway": "", "interface": ""}, "ipv6": {"gateway": "", "interface": ""}}
```

sonicaj:
@joachimmathes to confirm, is the PR up for review ?

joachimmathes:
@sonicaj Tests passed :tada:. I think it is.

iocage_lib/ioc_start.py
```diff
@@ -1312,19 +1311,27 @@ def __check_rtsold__(self):
...
-            )
+                silent=self.silent)
+
+    def get_default_gateway(self, address_family='ipv4'):
+        if self.host_gateways[address_family]['gateway']:
+            return re.sub(
+                '%.*$', '%epair0b',
+                self.host_gateways[address_family]['gateway'])
+        else:
+            return None
```
sonicaj: `None` as in the string `none` not python `None` ;) If you remember in an earlier PR you did we get the default gateway's value and if it's not `none` ( str one ), we set it else we don't. So this will fail there

iocage_lib/ioc_start.py
sonicaj: I think it might make more sense to replace it the following way then
1) `inet0` -> `epair0b`
2) `inet1` -> `epair1b`

iocage_lib/ioc_json.py
```diff
@@ -1207,6 +1213,12 @@ def check_jail_config(self, conf):
             # New jail creation
             pass
 
+        # Version 24 keys
```
sonicaj: Version `26` ;)
Also i don't like this personally for the reasons you pointed out earlier ( user explicitly setting this and on another bump this gets removed again resulting in more confusion ) but we will stick with it for now. I'll give higher priority to #963 now.

sonicaj (commented):
This is good, thanks again ;) Just a couple of minor notes left now

joachimmathes:
Thanks for the review. I will take care of that.

iocage_lib/ioc_start.py
joachimmathes: Oh yes: https://github.com/iocage/iocage/blob/43ba3fb55e557a5afcbb2a3bfb4288152ff04e49/iocage_lib/ioc_start.py#L955
I lost the big picture.

joachimmathes:
I've also missed to adjust https://github.com/iocage/iocage/blob/43ba3fb55e557a5afcbb2a3bfb4288152ff04e49/iocage_lib/ioc_start.py#L219 and https://github.com/iocage/iocage/blob/43ba3fb55e557a5afcbb2a3bfb4288152ff04e49/iocage_lib/ioc_start.py#L224.
But then again I am not happy with replacing constant `netifaces.AF_INET` by string `ipv4` and `netifaces.AF_INTE6` by `ipv6`. Because, from a coding point of view this is always error-prone. What do you think?

sonicaj:
Well we are not using `netifaces` anymore for this, so i believe we are good with `ipv4/6` constants ;)

joachimmathes:
I might have introduced a new constant like `ioc_common.AF_INET`, but I'm fine with the strings. :slightly_smiling_face:

joachimmathes:
After talking to my colleagues and thinking about it, I realized that mapping
1. `inet0 -> epair0b`
2. `inet1 -> epair1b`

is probably not what is intended. Finding the appropriate `epair` interface for a link local routing address depends on the bridge the `epair (vnet)` is connected to. I've provided some testcases to make things clearer.

iocage_lib/ioc_common.py
```diff
@@ -1050,3 +1050,31 @@ def parse_package_name(pkg):
         'revision': revision,
         'epoch': epoch,
     }
+
+
+def get_host_gateways():
+    gateways = {'ipv4': {'gateway': 'none', 'interface': 'none'},
```
sonicaj: Let's please stick to `None` here ( not the string ) as it was before. That is way easier to check in python and much neater imho. We can just return `"none"` from `default_gateway` function if no gateway is present.

iocage_lib/ioc_start.py
```diff
@@ -213,17 +214,15 @@ def __start_jail__(self):
             }, _callback=self.callback,
                 silent=self.silent)
 
-        if self.conf['vnet'] and self.defaultrouter == 'none':
-            self.log.debug('Grabbing default route')
-            self.defaultrouter = self.get_default_gateway()[0]
-            self.log.debug(f'Default Gateway: {self.defaultrouter}')
+        if self.conf['vnet'] and self.defaultrouter == 'auto':
+            self.log.debug('Grabbing IPv4 default route')
+            self.defaultrouter = self.get_default_gateway(netifaces.AF_INET)
```
sonicaj: You forgot to remove `netifaces.AF_INET` ;)

iocage_lib/ioc_start.py
```diff
@@ -213,17 +214,15 @@ def __start_jail__(self):
...
 
-            iocage_lib.ioc_common.logit({
-                'level': 'WARNING',
-                'message': f'{self.uuid}: vnet requires defaultrouter,'
-                           f' using {self.defaultrouter}'
-            }, _callback=self.callback,
-                silent=self.silent)
+        if self.conf['vnet'] and self.defaultrouter6 == 'auto':
+            self.log.debug('Grabbing IPv6 default route')
+            self.defaultrouter6 = self.get_default_gateway(netifaces.AF_INET6)
```
sonicaj: Same here

iocage_lib/ioc_start.py
```diff
@@ -1312,19 +1311,49 @@ def __check_rtsold__(self):
...
+            interface_bridge_map = self.get_interface_bridge_map()
+            ipv6 = gateway.split('%')[0]
+            if interface in interface_bridge_map:
+                jail_nic = interface_bridge_map[interface][0]
+            else:
+                iocage_lib.ioc_common.logit(
+                    {
+                        'level': 'EXCEPTION',
+                        'message': (f'No bridge for interface {interface}'
+                                    f'found in configuration.')
```
sonicaj: This is not a `f-string` so the prefixing `f` can be removed

iocage_lib/ioc_start.py
```diff
@@ -1312,19 +1311,49 @@ def __check_rtsold__(self):
...
+                    },
+                    _callback=self.callback,
+                    silent=self.silent)
+            if 'vnet' in jail_nic:
+                jail_gateway = f'{ipv6}%{jail_nic.replace("vnet", "epair")}b'
+            return jail_gateway
```
sonicaj: `jail_gateway` won't exist if `vnet` is not present in `jail_nic`

sonicaj:
@joachimmathes i wanted to ask do you use `accept_rtadv` for configuring your ip6_addresses with iocage and jails ?

joachimmathes:
@sonicaj No, we don't use `accept_rtadv`. Our iocage jail network setup is very explicit.

iocage_lib/ioc_start.py
```diff
@@ -1312,19 +1311,49 @@ def __check_rtsold__(self):
             stdout=su.PIPE
         )
 
-    def get_default_gateway(self):
-        # e.g response - ('192.168.122.1', 'lagg0')
-        try:
-            return netifaces.gateways()["default"][netifaces.AF_INET]
-        except KeyError:
+    def get_default_interface(self):
+        if self.host_gateways['ipv4']['interface'] != 'none':
```
sonicaj: @joachimmathes you forgot to update this usage ;) It will never be `none` (str)

sonicaj (commented):
@joachimmathes thank you so much for the updated changes ;) You just forgot to update one little teensy usage ;)

sonicaj:
> @sonicaj No, we don't use `accept_rtadv`. Our iocage jail network setup is very explicit.

Well i was hoping you might cause i noticed another issue and was wondering if you had encountered it before, but thanks!

iocage_lib/ioc_start.py
joachimmathes: I'm back from vacation. Will take care of that tomorrow.

joachimmathes: Fixed.

sonicaj (approved):
@joachimmathes this looks good to me, i have run some tests and everything seems good. Thank you for the contribution. I wanted to confirm this is fulfilling your use case ?

joachimmathes:
@sonicaj I've tested this PR in our IPv6-only environment once again. Everything worked as expected. So, I think we are done here. :smiley:

dlangille:
I think this pull has broken iocage for me today.  see https://github.com/iocage/iocage/issues/952#issuecomment-519186804

iocage_lib/ioc_start.py
```diff
@@ -1312,19 +1311,49 @@ def __check_rtsold__(self):
...
-            )
+                silent=self.silent)
+
+    def get_default_gateway(self, address_family='ipv4'):
+        gateway = self.host_gateways[address_family]['gateway']
+        interface = self.host_gateways[address_family]['interface']
+        if gateway and '%' in gateway:
+            interface_bridge_map = self.get_interface_bridge_map()
+            ipv6 = gateway.split('%')[0]
+            if interface in interface_bridge_map:
```
sonicaj, line 1334: @dlangille your issue seems to stem from this. As i do not have an ipv6 environment right now, would it be possible if you could test a patch for me ? I can bring up a PR and you can checkout to that branch and build iocage ?
Also @joachimmathes i may need to get in touch with you again to smooth this as we certainly are having regressions for ipv6 users making sure our new fixes don't regress for you. ( the issue @joachimmathes seems to stem from the fact that this change - `if gateway and '%' in gateway` - requires updates to the config which assume the user should know beforehand which bridge is going to be used. We most certainly do not want that as the idea is to make the process easier for the users. Btw can we please go through why we need to do this at all ? It seems it was working all right for @dlangille before and you yourself as well for other ipv6 use cases you have. Again apologies for a certain lack of knowledge here primarily because i can't test this myself right now ). Looking forward to hearing from you. Thank you!

dlangille: I only just saw this.  For those following along at home, we are athttps://github.com/iocage/iocage/pull/1016

dlangille: Thank you for your help on this. I have a working jail now.

PR #967, opened 2019-06-26
NAS-102274 / 11.3 / Update default json file

sonicaj:
This commit fixes a regression when which was introduced when we brought in plugin_name as a key in json configuration. The default json file was not updated previously which resulted in non plugin jails not being able to retrieve that prop at all.

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -700,7 +700,7 @@ class IOCConfiguration(IOCZFS):
     @staticmethod
     def get_version():
         """Sets the iocage configuration version."""
-        version = '24'
+        version = '24.1'
 
         return version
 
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_json.py
```diff
@@ -1071,6 +1071,10 @@ def check_config(self, conf, default=False):
         if not conf.get('nat_forwards'):
             conf['nat_forwards'] = 'none'
 
+        # Version 24 key
```
skarekrow, line 1074: 24.1

skarekrow (changes requested):

iocage_lib/ioc_json.py
sonicaj: It actually is 24's key :D

skarekrow (approved):

PR #969, opened 2019-06-27
NAS-102294 / 11.3 / Bug fix for listing admin portal

sonicaj:
This commit fixes a bug where if multiple ip's were specified for a plugin jail, it didn't list the admin portal correctly assuming it was only a single ip.

### DIFF
--- a/iocage_lib/ioc_list.py
+++ b/iocage_lib/ioc_list.py
@@ -339,14 +339,22 @@ class IOCList(object):
 
                 try:
                     with open(f"{mountpoint}/plugin/ui.json", "r") as u:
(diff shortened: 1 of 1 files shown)


skarekrow:
That’s purposely done behavior ;) as a lot of plugins/services aren’t listening on all ips but often just the first.

sonicaj:
That is true but it gives wrong results ;)
Like `http://192.168.0.24,192.168.0.25:8888/"`.
Also how do we identify that the other ip's aren't really being used by the service ?
IMO i think if the user specified multiple ip's, he/she should be responsible for identifying the correct one

skarekrow (approved):
Fair enough :)

PR #970, opened 2019-06-27
NAS-102297 / 11.3 / Bug fix for multiple ip's in vnet

sonicaj:
This commit fixes a bug where multiple ip's donot work with vnet on.

### DIFF
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -1181,7 +1181,7 @@ class IOCStart(object):
         if ipv6:
             ifconfig = [iface, 'inet6', ip, 'up']
         else:
-            ifconfig = [iface, ip, 'up']
+            ifconfig = [iface, ip, 'alias']
 
         try:
             if not wants_dhcp and ip != 'accept_rtadv':


iocage_lib/ioc_start.py
```diff
@@ -1181,7 +1181,7 @@ def start_network_vnet_addr(self, iface, ip, defaultgw, ipv6=False):
         if ipv6:
             ifconfig = [iface, 'inet6', ip, 'up']
         else:
-            ifconfig = [iface, ip, 'up']
```
skarekrow, line 1184: We need to only alias on VNET when the initial IP has been set. We don't want them all to be aliases

skarekrow (changes requested):

iocage_lib/ioc_start.py
william-gr: What do you mean, everything is an alias. There is no such thing as primary IP

iocage_lib/ioc_start.py
```diff
@@ -1181,7 +1181,7 @@ def start_network_vnet_addr(self, iface, ip, defaultgw, ipv6=False):
...
+            ifconfig = [iface, ip, 'alias']
```
william-gr, line 1184: You still need the up and the end

sonicaj: Nope. Docs say that when ever an ip is added, the interface is automatically changed to `up` state

william-gr (approved):

iocage_lib/ioc_start.py
skarekrow: As far as I know there is a difference. Aliases have a broadcast on just their IP instead of the whole subnet.

sonicaj: @skarekrow i haven't found anything which supports that ?
```
inet 192.168.0.26 netmask 0xffffff00 broadcast 192.168.0.255
net 192.168.0.27 netmask 0xffffff00 broadcast 192.168.0.255
```
Please advise. Also i am not sure what benefit we would have of having broadcast at all in that case ?

skarekrow (approved):

PR #984, opened 2019-07-05
NAS-102428 / 11.3 / Bug fixes while creating a plugin

sonicaj:
This commit fixes following bugs:
1) Correctly write the name of the plugin json file in jail.
2) Make sure IOCAGE_PLUGIN_IP is set in the environment correctly.
3) Admin portal would be misconfigured if multiple ip's were provided.

### DIFF
--- a/iocage_lib/ioc_plugin.py
+++ b/iocage_lib/ioc_plugin.py
@@ -653,21 +653,20 @@ fingerprint: {fingerprint}
         """Fetches the users artifact and runs the post install"""
         iocage_lib.ioc_start.IOCStart(self.jail, jaildir, silent=True)
 
-        ip4 = _conf["ip4_addr"]
-        if '|' in ip4:
-            ip4 = ip4.split("|")[1].rsplit("/")[0]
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_plugin.py
```diff
@@ -689,14 +688,16 @@ def __fetch_plugin_post_install__(self, conf, _conf, jaildir):
                     ], stdout=su.PIPE).stdout
                 )['jail-information']['jail'][0]['ipv4']
 
-            os.environ["IOCAGE_PLUGIN_IP"] = ip
+        self.log.debug(f'IP for plugin jail is {ip}.')
```
skarekrow:
```suggestion
        self.log.debug(f'IP for {self.plugin} - {self.jail}: {ip}.')
```

skarekrow (changes requested):

skarekrow (approved):
LGTM, we can probably drop the period from the debug log at the end. But that’s just being pedantic so up to you ;)

PR #985, opened 2019-07-05
NAS-102412 / 11.3 / Expose new api for listing resources

sonicaj:
This PR aims to introduce new api for listing resources.


skarekrow:
Let's not do this in master but a separate branch, since we haven't tagged for RELEASE ;)

sonicaj:
It does not affect the current interface/api at all ( almost ) ;) I think we are safe to keep it here. Thoughts ?

skarekrow:
Changing the interfaces at all this late into the RC is a Bad Idea™, I still think this should be a different branch. This is far more invasive then we want at this stage. We need to tag ;)

iocage_lib/listable_resources.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def __iter__(self):
+        if self.dataset_path:
+            for release in self.zfs.zfs_get_dataset_and_dependents(
+                self.dataset_path, depth=1
+            ):
+                yield self.resource(release)
+
+
+class ListableReleases(ListableResource):
```
william-gr: Seems like this should be in release.py

iocage_lib/listable_resources.py
```diff
@@ -0,0 +1,68 @@
...
+import os
+import re
+import requests
+
+from iocage_lib.ioc_fetch import IOCFetch
+from iocage_lib.ioc_json import IOCZFS
+from iocage_lib.release import Release
+
+
+class ListableResource(collections.abc.Iterable):
```
william-gr: And this should be in resource.py? Why separate them

sonicaj: Well my thinking was having all resources which can be listed in a separate file. If you think it makes more sense to have these in `resource.py`, i'll update accordingly. ( i didn't want to populate a single namespace too much )

william-gr: Its easier to use if they are in the same, they are all about the same thing, resources (listing and the resource itself).

william-gr (approved):

skarekrow (dismissed):
Branch is still my issue

william-gr:
Why its an issue? It doesnt change the current API.

skarekrow:
> Why its an issue? It doesnt change the current API.

It changes a few areas in ioc_json and the whole point of the RC is to stabilize and release ;) Branch is a better place as we are waiting on a different or to fix ipv6 before we tag. No point adding or changingnthese interfaces before then, especially this close.

iocage_lib/release.py
```diff
@@ -0,0 +1,56 @@
...
+        if eol_check and remote:
+            # TODO: Please let's not use this in the future and look at
+            # comments above
+            self.eol_list = IOCFetch.__fetch_eol_check__()
+
+    def __iter__(self):
+        if self.remote:
+            # TODO: Please abstract this in the future
+            req = requests.get(
+                'https://download.freebsd.org/ftp/'
```
skarekrow, line 41: Isn’t valid for HBSD we should skip this for that

william-gr:
When are you going to release? Should we put this PR in freenas/iocage instead?

skarekrow:
> When are you going to release? Should we put this PR in freenas/iocage instead?

Why? Neither are my choice. Waqar is in charge of this release, and the PR were waiting on is a community PR.

iocage_lib/release.py
sonicaj: @skarekrow I don't want to overload this with checks or improve things which we are going to remove in the future ( double effort ;) ). Moving on i am going to remove this and abstract host/distribution which would cover this. So this is not needed for now and this api is incompatible with HBSD until we abstract this even more.

sonicaj: In other words a WIP ;)

PR #990, opened 2019-07-11
NAS-102494 / 11.3 / Allow lzma compression for importing/exporting jails

sonicaj:
This PR aims to expose option to allow lzma compression for import/export of jails.

### DIFF
--- a/iocage_lib/iocage.py
+++ b/iocage_lib/iocage.py
@@ -987,7 +987,7 @@ class IOCage(ioc_json.IOCZFS):
                     _callback=self.callback,
                     silent=self.silent)
 
-    def export(self, compression_algo):
+    def export(self, compression_algo='zip'):
         """Will export a jail"""
         uuid, path = self.__check_jail_existence__()
         status, _ = self.list("jid", uuid=uuid)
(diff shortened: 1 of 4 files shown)


iocage_lib/iocage.py
```diff
@@ -1386,9 +1388,11 @@ def get(
 
             return jail_list
 
-    def import_(self):
+    def import_(self, compression_algo='zip'):
```
william-gr: Is there a way to autodetect the file instead?

sonicaj: Well we can but that means an exception if we have 2 exported jails with similar names but different algorithms. I guess i raise an exception in that case and that will allow this to automatically proceed if there are no conflicts. Thoughts ?

william-gr (commented):
I find it weird that we cannot provide a path of the image we want to import?

sonicaj:
Well we have a defined dataset for it right now where all images must live. But for import cases it indeed is. Would you like me to add an option ?

william-gr:
Yes, also seems easier for use case instead of having to move it first to images dataset.

william-gr (approved):

PR #998, opened 2019-07-23
NAS-102598 / 11.3 / Introduce pre/post update hooks

sonicaj:
1) Introduce a pre update hook which can be executed before a plugin is updated.
2) Update existing post update hook to have it's script renamed to `post_update.sh` instead of `post_upgrade.sh` as the former correctly represents the operation in progress.

### DIFF
--- a/iocage_lib/ioc_plugin.py
+++ b/iocage_lib/ioc_plugin.py
@@ -941,6 +941,45 @@ fingerprint: {fingerprint}
 
         return plugin
 
+    def __run_hook_script__(self, script_path):
+        self.__stop_rc__()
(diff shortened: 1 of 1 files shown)


themylogin (approved):

skarekrow:
`post_upgrade` *is* actually ran on upgrade as well :) You'll need to handle any plugins that currently use that, as this is a breaking change. I suggest looking for either file.

sonicaj:
@skarekrow there is only one plugin using this `mineos` an we are in the process of changing this there ;)
Plugin update method is executed on plugin upgrade ;) But still update refers to this correctly imo :)

PR #1002, opened 2019-07-26
NAS-102662 / 11.3 / Disallow running jails using same system ports

sonicaj:
This commit disallows running a jail with NAT which wants to consume a system port which is already in use by another jail/entity.

### DIFF
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -1078,3 +1078,21 @@ def get_host_gateways():
             gateways[af_mapping[af]]['interface'] = \
                 default_route[0]['interface-name']
     return gateways
+
+
+def get_jails_with_config(filters=None, mapping_func=None):
+    # FIXME: Due to how api is structured, there is no good place to put this
+    #  so when we move on with restructuring the api, let's remove this as well
+    #  importing iocage_lib.iocage above gives us a circular dep due to how
+    #  iocage designates iocage_lib.iocage at top and imports everything else
+    #  within.
+    import iocage_lib.iocage
+    return {
+        j['host_hostuuid']: j if not mapping_func else mapping_func(j)
+        for j in map(
+            lambda v: list(v.values())[0],
+            iocage_lib.iocage.IOCage(jail=None).get(
+                'all', recursive=True
+            )
+        ) if not filters or filters(j)
+    }
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -26,6 +26,7 @@ import datetime
 import hashlib
 import os
 import re
+import itertools
 import shutil
 import json
 import subprocess as su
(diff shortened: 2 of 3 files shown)


iocage_lib/ioc_common.py
```diff
@@ -1078,3 +1078,42 @@ def get_host_gateways():
             gateways[af_mapping[af]]['interface'] = \
                 default_route[0]['interface-name']
     return gateways
+
+
+def parse_nat_fwds(forwards, log=None):
```
skarekrow: Passing log makes no sense, this isn't a common operation. This is exclusive to nat, it belongs in that module.

iocage_lib/ioc_common.py
```diff
@@ -1078,3 +1078,42 @@ def get_host_gateways():
...
+        except ValueError:
+            map = port
+
+        if log:
+            log.debug(f'Mapping {port} to {map}')
+
+        yield proto, port, map
+
+
+def get_jails_with_config(filters=None, mapping_func=None):
```
skarekrow, line 1083: Why do this instead of just using ioc_json? That's the good place to put this :P

iocage_lib/ioc_start.py
```diff
@@ -202,6 +204,34 @@ def __start_jail__(self):
...
+                            j['nat_forwards']
+                        )
+                    ]
+                ).values(), self.blacklist_ports
+            )):
+                iocage_lib.ioc_common.logit(
+                    {
+                        'level': 'EXCEPTION',
+                        'message': f'Please correct {nat_forwards} port rule '
+                        'as another already running jail is using one of the '
```
```suggestion
                        'as another running jail is using one of the '
```

iocage_lib/iocage.py
```diff
@@ -1880,7 +1880,7 @@ def __soft_restart__(self):
                 _callback=self.callback,
                 silent=self.silent)
 
-    def start(self, jail=None, ignore_exception=False):
+    def start(self, jail=None, ignore_exception=False, blacklist_ports=None):
```
skarekrow: Perhaps `used_ports` explains what it does better?

skarekrow (changes requested):

iocage_lib/ioc_common.py
sonicaj: That would mean getting list of jails which exist and then retrieving their config, why duplicate code when we already are doing that in iocage_lib.iocage ?

skarekrow: We're duplicating code here, the lib's usage is a small wrapper. It helps avoid a recursive dep

sonicaj: I am sorry but that solution is way more bad then this one. We would need to use iocage.list to first retrieve list of datasets and once being done with that retrieve props not to mention corrupt jails. I would rather personally stick to a recursive dep :p Essentially the whole functionality of iocage.list + iocage.get ( all, non recursive ) - a good number of lines. Please advise @skarekrow @william-gr .

skarekrow: Heh, that's fine then. It will be restructured eventually anyways.

sonicaj: Yup ;)

skarekrow (approved):

PR #1012, opened 2019-08-04
NAS-102771 / 11.3 / Improve validation for nat default props

sonicaj:
This commit adds validation to prevent users from changing default nat props like nat_prefix. Reasoning behind this is that the pf/ipfw rules don't anticipate this change and hence fail to work as desired for jails which are already running with the old prefix.

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -2323,14 +2323,31 @@ class IOCJson(IOCConfiguration):
 
         if key in (
             'nat_prefix', 'nat_interface', 'nat_backend'
-        ) and not default:
-            iocage_lib.ioc_common.logit(
-                {
-                    'level': 'EXCEPTION',
-                    'message': f'{key} can only be changed for defaults!'
-                },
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_json.py
```diff
@@ -2323,14 +2323,26 @@ def json_check_prop(self, key, value, conf, default=False):
...
+                    silent=self.silent)
+
+            if iocage_lib.ioc_common.get_jails_with_config(
+                lambda j: (j['state'] == 'up' and j['nat'])
+            ):
+                iocage_lib.ioc_common.logit(
+                    {
+                        'level': 'EXCEPTION',
+                        'message': f'{key} cannot be changed with active '
+                        'NAT jails. Please stop them first.'
```
skarekrow: We should list which jails are active I think.

skarekrow (changes requested):

skarekrow (approved):

skarekrow:
LGTM Thanks! 👍

PR #1014, opened 2019-08-05
NAS-102750 / 11.3 / Update basejails configuration correctly

sonicaj:
This commit makes sure we update basejails configuration correctly. If a basejail is not running, freebsd-version file will be non-existent and that would result in a corrupt configuration for the basejail in question as it won't be updated.

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -37,6 +37,7 @@ import sys
 import iocage_lib.ioc_common
 import iocage_lib.ioc_create
 import iocage_lib.ioc_exec
+import iocage_lib.ioc_fstab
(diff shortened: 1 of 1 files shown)


skarekrow (commented):
Why not use the jails RELEASE prop?

sonicaj:
The idea is to correct the value of RELEASE prop ;)

sonicaj:
In our earlier versions, we had a few bugs which resulted in misconfigured release property, so we introduced these checks to ensure that each time a config is updated, RELEASE prop reflects the correct release version.

skarekrow:
Whoops forgot that! Haha

skarekrow (approved):

PR #1021, opened 2019-08-14
Correctly set scope id in ipv6

sonicaj:
This commit makes sure we correctly set ipv6 scope id in default route if explicitly present in default route.

### DIFF
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -994,7 +994,9 @@ class IOCStart(object):
                     lambda v: v[0] != 'none' and v[1] != 'none',
                     net_configs
                 ):
-                    if ipv6:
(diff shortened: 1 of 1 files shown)


sonicaj:
@sahlawat can you please confirm if this fixes your issue ?

sahlawat:
Was able to successfully create and destroy a jail with IPv4 and IPv6 network configurations using the "ipv6" branch. Thank you.

sonicaj:
I assume it’s safe to say it started okay too ?

sahlawat:
sorry :) - yes in between the create and destroy I logged in and verified IPv4 and IPv6 reachability and am also able to stop and start my other jails without issues. Will continue using this branch till this change merges into master.

skarekrow (approved):

Issue #992, opened 2019-07-15
iocage upgrade incorrectly reports as failed upgrading from 11.2-RELEASE to 11.3-RELEASE


vendion:
Make sure to follow and check these boxes before submitting an issue! Thank you.

- [x] Supply `iocage --version`
  Version	1.1 RELEASE 2019/01
- [x] Supply the commands used, along with any steps to recreate it.
  `# iocage upgrade -r 11.3-RELEASE test`
- [x] Provide the output from the command you issued.
```
To install the downloaded upgrades, run "/tmp/tmplf8aa2n_ install".
Installing updates...
Kernel updates have been installed.  Please reboot and run
"/tmp/tmplf8aa2n_ install" again to finish installing updates.
Installing updates... done.
No updates are available to install.
Run '/tmp/tmplf8aa2n_ fetch' first.
Command: /tmp/tmplf8aa2n_ -b /iocage/jails/test/root -d /iocage/jails/test/root/var/db/freebsd-update/ -f /iocage/jails/test/root/etc/freebsd-update.conf -r 11.3-RELEASE install failed!
# iocage console test
(inside upgraded jail) # freebsd-version
11.3-RELEASE
```
- [x] Supply what you expected the result or output to be
   No error message stating that the upgrade has failed
- [x] Checked that the problem has not already been fixed on `master` if using
 a stable release.
   I don't see any closed tickets regarding this, may have missed one though

Even though when inside the Jail `freebsd-version` reports that it is 11.3-RELEASE, iocage still shows the jail is 11,2-RELEASE 

```
# iocage get release test
11.2-RELEASE-p11
```

adam12b1:
We are seeing the same problem. The upgrade process always dies on the same generic error, when trying to upgrade a single standalone jail with `iocage upgrade -r 11.3-RELEASE [jailname]`:
```
Command: /tmp/tmplf8aa2n_ -b /iocage/jails/test/root -d /iocage/jails/test/root/var/db/freebsd-update/ -f /iocage/jails/test/root/etc/freebsd-update.conf -r 11.3-RELEASE install failed!
```
We're also having a problem when trying to use `iocage upgrade ALL` (see #971 ), but that seems to be totally separate and much more specific (only about the ALL feature).

adam12b1:
Update on this issue: It appears to be specific to the upgrade from 11.2 to 11.3, and does _not_ occur when upgrading all the way from 11.2 to 12.0-RELEASE.

grembo:
[...]

PR #1027, opened 2019-08-20
Check if there are updates to install before invoking `freebsd-update install' on upgrade.

grembo:
This un-breaks updates to 11.3-RELEASE. First reported in
iocage/iocage#992.

Since r345055[0] (first used in 11.3-RELEASE, but not
12.0-RELEASE), freebsd-update returns 1 in case no updates are
available to install and fetch wasn't called first. This
restored the behaviour it had until r324441 and was required to
unbreak ezjail and iocell[1].

This changed caused iocage to assume the update failed and kept
it from updating config.json properly.

*Updated 2019-08-24:*
This change, which I also discussed with the original author of
freebsd-update, fixes this problem in a straight-forward
manner that is backwards compatible (using the symbolic
link is considered a stable enough interface to freebsd-update).

As a bonus, it doesn't rely on a hard coded number of invocations
of freebsd-update any more (which was fixed to "3" before).

As part of the update, I removed dead/unreachable exception
handling code that would've also done the wrong thing if it was
ever reached.

As part of testing this change I also tried running upgrade
with interactive set to False (manually in the source),
which also worked ok. It might make sense to remove non-interactive
code from ioc_upgrade in the future though.

[0]https://svnweb.freebsd.org/base?view=revision&revision=345055
[1]https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=229346

- [x] Explain the feature

### DIFF
--- a/iocage_lib/ioc_upgrade.py
+++ b/iocage_lib/ioc_upgrade.py
@@ -24,6 +24,7 @@
 """iocage upgrade module"""
 import datetime
 import fileinput
+import hashlib
 import os
 import pathlib
 import subprocess as su
@@ -80,6 +81,11 @@ class IOCUpgrade(iocage_lib.ioc_json.IOCZFS):
         }
 
         self.callback = callback
+
+        # symbolic link created on fetch by freebsd-update
+        bd_hash = hashlib.sha256((self.path + '\n').encode('utf-8')).hexdigest()
+        self.freebsd_install_link = os.path.join(self.path,
+            'var/db/freebsd-update', bd_hash + '-install')
 
     def upgrade_jail(self):
         tmp_dataset = self.zfs_get_dataset_name('/tmp')
@@ -140,44 +146,40 @@ class IOCUpgrade(iocage_lib.ioc_json.IOCZFS):
                         callback=self.callback
                     )
             else:
-                try:
-                    iocage_lib.ioc_exec.InteractiveExec(
-                        fetch_cmd,
-                        self.path.replace('/root', ''),
-                        uuid=self.uuid,
-                        unjailed=True
-                    )
-                except iocage_lib.ioc_exceptions.CommandFailed:
-                    self.__rollback_jail__()
-                    msg = f'Upgrade failed! Rolling back jail'
-                    iocage_lib.ioc_common.logit(
-                        {
-                            "level": "EXCEPTION",
-                            "message": msg
-                        },
-                        _callback=self.callback,
-                        silent=self.silent
-                    )
+                iocage_lib.ioc_exec.InteractiveExec(
+                    fetch_cmd,
+                    self.path.replace('/root', ''),
+                    uuid=self.uuid,
+                    unjailed=True
+                )
 
-            if not self.interactive:
-                while not self.__upgrade_install__(tmp.name):
-                    pass
-            else:
-                # FreeBSD update loops 3 times
-                for _ in range(3):
-                    try:
-                        self.__upgrade_install__(tmp.name)
-                    except iocage_lib.ioc_exceptions.CommandFailed:
-                        self.__rollback_jail__()
-                        msg = f'Upgrade failed! Rolling back jail'
-                        iocage_lib.ioc_common.logit(
-                            {
-                                'level': 'EXCEPTION',
-                                'message': msg
-                            },
-                            _callback=self.callback,
-                            silent=self.silent
-                        )
+            if not os.path.islink(self.freebsd_install_link):
+                msg = f'Upgrade failed, nothing to install after fetch!'
+                iocage_lib.ioc_common.logit(
+                    {
+                        'level': 'EXCEPTION',
+                        'message': msg
+                    },
+                    _callback=self.callback,
+                    silent=self.silent
+                )
+
+            for _ in range(50): # up to 50 invocations to prevent runaway
+                if os.path.islink(self.freebsd_install_link):
+                    self.__upgrade_install__(tmp.name)
+                else:
+                    break
+
+            if os.path.islink(self.freebsd_install_link):
+                msg = f'Upgrade failed, freebsd-update won\'t finish!'
+                iocage_lib.ioc_common.logit(
+                    {
+                        'level': 'EXCEPTION',
+                        'message': msg
+                    },
+                    _callback=self.callback,
+                    silent=self.silent
+                )
 
             new_release = iocage_lib.ioc_common.get_jail_freebsd_version(
                 self.path,
@@ -335,7 +337,7 @@ class IOCUpgrade(iocage_lib.ioc_json.IOCZFS):
         return new_release
 
     def __upgrade_install__(self, name):
-        """Installs the upgrade and returns the exit code."""
+        """Installs the upgrade."""
         install_cmd = [
             name, "-b", self.path, "-d",
             f"{self.path}/var/db/freebsd-update/", "-f",
@@ -351,16 +353,10 @@ class IOCUpgrade(iocage_lib.ioc_json.IOCZFS):
                 unjailed=True,
                 callback=self.callback,
             ) as _exec:
-                output = iocage_lib.ioc_common.consume_and_log(
+                iocage_lib.ioc_common.consume_and_log(
                     _exec,
                     callback=self.callback
                 )
-
-            for i in output['stdout']:
-                if i == 'No updates are available to install.':
-                    return True
-
-            return False
         else:
             iocage_lib.ioc_exec.InteractiveExec(
                 install_cmd,


iocage_lib/ioc_upgrade.py
```diff
@@ -112,7 +113,17 @@ def upgrade_jail(self):
...
-                tmp.write(fbsd_update.read())
+                # Since r345055 (11.3-RELEASE) freebsd-update
+                # returns 1 in case fetch needs to be called,
+                # this works around this problem.
+                exit_code_regex = \
+                    rb'''\t\t\techo "Run \'\$0 fetch\' first."\n'''\
+                    rb'''\t\t\texit 1'''
+                exit_code_sub = \
+                    rb'''\t\t\techo "Run '$0 fetch' first."\n'''\
+                    rb'''\t\t\t#exit 1 #iocage'''
```
skarekrow: Hmm this is a tough problem, because if any output on their end changes, this all breaks. Parsing userland output is suboptimal, but I can't think of a better solution either. Ideally they don't continually change this script for other programs, without thinking about other consumers of it.

iocage_lib/ioc_upgrade.py
```diff
@@ -140,44 +151,20 @@ def upgrade_jail(self):
...
                     )
             else:
-                try:
-                    iocage_lib.ioc_exec.InteractiveExec(
-                        fetch_cmd,
-                        self.path.replace('/root', ''),
-                        uuid=self.uuid,
-                        unjailed=True
-                    )
-                except iocage_lib.ioc_exceptions.CommandFailed:
```
skarekrow, line 150: My memory suggests we actually did snapshot the upgraded jails at one time, I can't seem to find where we do that anymore. But my suggestion here is instead of removing this code, that we actually just snapshot the jail if it's not a basejail (pluginv2 also) type. We want the upgrade to be reversible in case of failures after all.

skarekrow (changes requested):
Thanks for this! Just a suggestion

iocage_lib/ioc_upgrade.py
grembo: Well, they changed it once in 2017 and now changed it back :). I assume this will be relatively stable (and they way this is implemented it won't break). Ideally we could add an environment variable to freebsd-update to control this behaviour (I might create a patch let's see), but this won't help versions in existence (like 11.3-RELEASE). I was a bit surprised iocage fetches it from github though and not from a direct freebsd resource (given that patches are fetched from there, it would make sense to not introduce an additional resource). Outside of the scope of this patch, but still.

grembo: p.s. This is not parsing userland output, but modifying freebsd-update to not exit with this code. Parsing userland output would've been more unstable and in this case really hard (like parse it in python while still having full interactivity). And as this is basically just a search/replace, if the code changes in a compatible way, it will simply be a no-op.

skarekrow: Something in that script we can call in the environment would be more ideal I think. As far as fetching from github, it was simply because I couldn't find a mirror of freebsd-update that would get the minor patches as the major lived. 

I'm open to changing that behavior of course! So with your suggestion I think we should do the following:
- Use an env variable or flag to freebsd-update to *not* exit 1 on these things
- Only run this sub on 11.3-RELEASE and under.

That should help us with stability in regards to this behavior as those shouldn't change at all.

iocage_lib/ioc_upgrade.py
grembo: Jails are running during the upgrade, so if anything stateful happens during the upgrade (e.g. file upload, database change, or even just a log file entry) an automatic rollback would be disastrous for the user. I assume this is the reason why automatic snapshots/rollbacks exist for basejails (stateless) and not for jails (stateful).

As a side note, iocage is starting a jail silently (and without interacting with the user) before upgrading it. As a new user who stopped the jail before running upgrade and then finding it running during the process, this came as a bit of a surprise.

iocage_lib/ioc_upgrade.py
skarekrow: > p.s. This is not parsing userland output, but modifying freebsd-update to not exit with this code. Parsing userland output would've been more unstable and in this case really hard (like parse it in python while still having full interactivity). And as this is basically just a replace, if the code changes in a compatible way, it will simply be a no-op.

Yeah I noticed I made a mistake and this is actually the scripts content but I didn't want to edit it and make any reply look out of context :D

iocage_lib/ioc_upgrade.py
skarekrow: The jail should be stopped by iocage, snapshotted, started, upgraded (anything during this time should be considered expendable, as if the upgrade fails, the user would likely prefer the state their jail was at *before* any changes happened) and then if success destroy snapshot.

As for it being a surprise, I understand. We need the mounts, so it's the best we got. I suppose some documentation or a print would help a user understand that clearer.

grembo: > the user would likely prefer the state their jail was at _before_ any changes happened) and then if success destroy snapshot.

I wouldn't make that assumption, as this really depends on the use case. If the jail does anything stateful that has any sort of meaning (e.g. communicating with the outside world), I absolutely prefer to not have the state that was accumulated during an upgrade destroyed. Also if my jail setup is a bit more complicated (think ZFS mountpoints from outside the main jails containers) it would end up in quite a mess. Given that upgrades can easily take 15 minutes and longer, and that the jail is started with all services running for update (correct me if I'm wrong here), that is a lot of state to be lost. So, unless we have an enforced setup, in which the jail itself is known to be stateless - like an ideal container, where all logging and all storage lives outside of the jail - automatic rollback is asking for disaster and should be an optional feature.

skarekrow: Forgive me as it's been a while since I've played around with upgrade, as @sonicaj has taken more of the direct hands on with the project lately.

I think that's a great point. Thinking about this, rollback should be an optional flag on upgrade, but in addition we should start the jail process and not the rc process, so we can ensure these things aren't happening during upgrade. You're right, this process can take quite some time depending on a variety of factors.

iocage_lib/ioc_upgrade.py
grembo: I created a patch to freebsd-update and opened a bug:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=239997

If approved I would modify this pull request to make use of it.

skarekrow: Thanks for all the effort @grembo 👍

grembo: @skarekrow After discussing the issue with @cperciva, I took his advice and implemented a different fix, that doesn't require patching freebsd-update on download and is backwards compatible. It looks for the symlink created by freebsd-update to determine if there is anything left to install (this is a stable enough interface to freebsd-update at this point). This also has the advantage of replacing the for loop hardcoded to three iterations with a while condition and also removing error messages from the log. I'll still look into patching freebsd-update to add a function called *freebsd-update updatesready*, but more as a new feature and not as a pre-condition to fix iocage. I feel like getting this in (also in the ports tree as a patch to 1.1) should be a priority, as it's actually breaking people's setup on upgrade to 11.3 at the moment.

skarekrow: Agreed, this is something that has been a pain point for a lot of folks recently.

skarekrow (dismissed):
LGTM @sonicaj over to you

iocage_lib/ioc_upgrade.py
```diff
@@ -140,44 +146,26 @@ def upgrade_jail(self):
...
-                        iocage_lib.ioc_common.logit(
-                            {
-                                'level': 'EXCEPTION',
-                                'message': msg
-                            },
-                            _callback=self.callback,
-                            silent=self.silent
-                        )
+            while os.path.islink(self.freebsd_install_link):
+                self.__upgrade_install__(tmp.name)
```
sonicaj: Is it possible that this never ends ?

sonicaj (commented):
Thank you for the contribution @grembo. It's much appreciated. I had a quick query, have you tested this across different releases i.e 11.2->12 / 11.3->12 / 11.1 -> 11.2 etc ?
I just want to be sure that this behavior does not disrupt other upgrade paths ( it looks like it shouldn't but confirming it with you - also you do mention that it is backwards compatible but just being sure ;) )

iocage_lib/ioc_upgrade.py
grembo: No, there is no code path in freebsd-update install that would exit on 0 and leave ${BDHASH}-install behind.

cperciva: There is, but only after `freebsd-update upgrade` when it wants you to reboot after it installs a new kernel / before it deletes old shared libraries.

grembo:
> Thank you for the contribution @grembo. It's much appreciated. I had a quick query, have you tested this across different releases i.e 11.2->12 / 11.3->12 / 11.1 -> 11.2 etc ?
> I just want to be sure that this behavior does not disrupt other upgrade paths ( it looks like it shouldn't but confirming it with you - also you do mention that it is backwards compatible but just being sure ;) )

Okay, I did *a lot* of testing - unfortunately freebsd-update isn't super fast and debugging the issues found took some time as well.

As expected, the patch works okay on all upgrade paths I tested, which were:

- 11.1 -> 11.2
- 11.2 -> 11.3
- 11.1 -> 12.0
- 11.2 -> 12.0
- 11.3 -> 12.0
- Multiple upgrades (iocage upgrade ALL -r 11.3/12.0)

I found few other problems that way though - all of which are unrelated to the patch (I tested with a patched version of iocage 1.1):

1. When upgrading a jail to 11.x on a 12.0 jailhost, all config merging breaks/creates empty merged files. This is quite annoying, as every file that had changes/required merge comes up in $EDITOR as an empty file and, if the process isn't cancelled, will replace the configuration file in question with that empty "merged file". This is due to `merge` not being available on 12 (rcs was removed from base and 12.x freebsd-update uses `diff3` instead). I fixed that by adding devel/rcs57 as a run time dependency in portrevision 1 (based on looking for `merge`).
2. The UNAME_r workaround in ioc_upgrade.py 1.1 breaks multiple upgrades (like in `iocage upgrade ALL -r ...` as it affects os.name) - I added a patch for that to portrevision 1, as the code it worked around isn't in any release and was fixed (it's also gone in iocage master already)
3. When upgrading to anything released before 11.3 (that includes 12.0), the source component is never updated due to a bug in freebsd-update which was only fixed in 11.3-RELEASE - I didn't address this yet, as I feel like it would further delay the more urgent fixes (I think this will require some patching of freebsd-update, like I did in the first version of *this* patch). I'll open a separate issue and/or pull request to fix this once this pull request is done.

You can find the bug and patches to the port skeleton of sysutils/iocage here:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=240177

iocage_lib/ioc_upgrade.py
grembo: > There is, but only after `freebsd-update upgrade` when it wants you to reboot after it installs a new kernel / before it deletes old shared libraries.

Touché - I clearly wasn't precise enough, sorry. What I meant to say is, that freebsd-update install either moves the installation forward or exits != 0. It doesn't exit on 0 and gets stuck in whatever it was doing beforehand. You can trust its exit code - it won't say "ok" if there is a problem.

Edit: At first I thought of having a counter in this code to prevent a runaway, but looking at freebsd-update it felt that all it did was making the code harder to read.

The (simplified) logic is basically:
- Check params - if there's a problem, exit 1 (unless ${BDHASH}-install doesn't exist and fetch was called, in which case it exits 0 - but this code path in unreachable in our case)
- install_files:
  + If !kernel_done: install kernel, exit 1 on error. touch kernel_done; exit 0
  + install other files - exit 1 on any error
  + rename ${BDHASH}-install to ${BDHASH}-rollback

grembo: Quite silly of me, as actually the whole point of this patch is to look if ${BDHASH}-install was left behind while `freebsd-update install` exited on 0 to know that additional invocations are required. ^_^

The point is, that, unless there are bugs introduced in future versions of freebsd-update, this loop should be safe by all means.

grembo: @sonicaj So the correct answer should've been:

"No, there is no code path in `freebsd-update install` that would exit on 0 and leave ${BDHASH}-install behind _without progressing the installation process_."

@cperciva Does that sound acceptable?

cperciva: Yes, that sounds right.

sonicaj: @grembo @cperciva thank you for clarifying. @skarekrow are you comfortable with this loop ? I am not sure how to proceed with this, if anything changes on freebsd-update end even as a bug for some time, this is going to repeat itself endlessly. If you think we are good, i am okay with it then.

sonicaj (commented):
@grembo thanks again for your contribution and for thoroughly testing the cases. I just have a minor query for @skarekrow, else we should be good.

iocage_lib/ioc_upgrade.py
sonicaj: @skarekrow i feel like maybe putting an upper limit here ? i.e 5 maybe ? not sure, i just don't want us to be affected by any unwanted upstream changes. @grembo would you have any leads ?

grembo: @sonicaj If it helps: If FreeBSD-update changed in the future, this would also affect ezjail and iocell (and probably many other handcrafted solutions). The advantage iocage has over ezjail and iocell with this patch is that it can tell apart „finished“ and „something broke“ (they run “while ok”, while this code runs “while file is there, expect ok”). Besides, a looping upgrade process will certainly cause an issue being opened *immediately*.

In general, I’m all for defensive coding and my first instinct was to put in a runaway handler - just looking the at remainder of iocage I think it’s written to trust return codes from the various tools used.

That said, if these concerns delay the patch, I’m more than happy to add an upper limit + exception - I’m absolutely not religious about it. I would select a fairly high number like 50 iterations in this case though and fail the upgrade process once it’s reached.

skarekrow: I think it's reasonable for us to rely on exit codes of our upstream projects. Which we do in a lot of cases. I'm not against defensive programming either, but like @grembo mentioned, a user will very likely notice that the script is indeed looping and kill it, and (hopefully) promptly filing an issue.

What I want to avoid is us having to low of a ceiling in case upstream changes how things are done (perhaps it becomes a wrapper to pkg base in the future -- who knows), so I agree with @grembo's suggestion of 50. That seems rather high, but it's high enough we don't worry about prematurely killing things, and low enough that the program won't spin for ages.

Regardless, this essentially replaces what we did before which was a fairly naive loop that just hoped for the best. I don't think this PR will bring us any additional risk of undesired behavior, and the fix is worth the risk imo.

So tldr; let's add a 50 iteration ceiling, that's a good enough compromise.

sonicaj: @grembo can you please add a ceiling of 50 iterations ? Thank you
@skarekrow i agree with the compromise ;)

skarekrow (approved):
You rock @grembo, thanks again!

iocage_lib/ioc_upgrade.py
grembo: @sonicaj I implemented a runaway check, also fixed an indentation I got wrong in the previous version and corrected a function description I stumbled upon (__upgrade_install__ doesn't return anything any more and even before it didn't return the exit code, but True/False/None).

grembo:
> You rock @grembo, thanks again!

You're welcome :)

grembo:
I updated the patch to the port skeleton at https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=240177 to match the approved version of this pull request.

sonicaj (approved):
That is much better, thank you @grembo ; )

araujobsd:
I'm late to the party!!!! I have a bug report on FreeBSD made by @grembo, I'm wondering if we might soon have a new iocage release (it is time), or should I add the patch on the ports till we wait for a new iocage release that will incorporate this github pr. @skarekrow and @sonicaj any plan?

sonicaj:
@araujobsd it is long overdue. We have been looking at regressions from the recent changes ( nat etc ) and i think we are stable enough now. @skarekrow do you think we should perhaps cut one later this week or at the start of the next one ?

grembo:
> I'm late to the party!!!! I have a bug report on FreeBSD made by @grembo, I'm wondering if we might soon have a new iocage release (it is time), or should I add the patch on the ports till we wait for a new iocage release that will incorporate this github pr. @skarekrow and @sonicaj any plan?

@araujobsd Regardless of a release I think it would make sense to apply the patch to the port, as

  1. The dependency on rcs is important for 12-RELEASE users running 11-RELEASE jails (otherwise all config file merges fail on upgrade)
  2. We could also make it available to users of the quarterly branch - putting an upgrade to 1.2 in there would be bad style (too many changes), while adding a relevant bugfix in there is recommended
  3. It would give those struggling with updating to 11.3 an immediate solution, without having to upgrade to 1.2 (and without introducing an additional dependency on the release)

grembo:
> @araujobsd it is long overdue. We have been looking at regressions from the recent changes ( nat etc ) and i think we are stable enough now. @skarekrow do you think we should perhaps cut one later this week or at the start of the next one ?

@sonicaj There is another upgrade problem caused by freebsd-update not detecting the src component correctly if BASEDIR isn't / - so if the jailhost has no src tree, src trees in iocage jails will never be updated.

This seems to be a problem for quite some time now. It was attempted to patch this in freebsd-update earlier this year, but the patch didn't work due to the way options are evaluated in freebsd-update. I opened a [review](https://reviews.freebsd.org/D21473) to fix this in base, I'm curious if we should add something in iocage to address it though (as a user, my expectation of an iocage upgrade operation would be to end up with a _fully_ updated jail, including src).

sonicaj:
> > @araujobsd it is long overdue. We have been looking at regressions from the recent changes ( nat etc ) and i think we are stable enough now. @skarekrow do you think we should perhaps cut one later this week or at the start of the next one ?
> 
> @sonicaj There is another upgrade problem caused by freebsd-update not detecting the src component correctly if BASEDIR isn't / - so if the jailhost has no src tree, src trees in iocage jails will never be updated.
> 
> This seems to be a problem for quite some time now. It was attempted to patch this in freebsd-update earlier this year, but the patch didn't work due to the way options are evaluated in freebsd-update. I opened a [review](https://reviews.freebsd.org/D21473) to fix this in base, I'm curious if we should add something in iocage to address it though (as a user, my expectation of an iocage upgrade operation would be to end up with a _fully_ updated jail, including src).

Well @grembo, i have already submitted a patch in github and opened up an issue in bugzilla for it. There hasn't been any interest shown by freebsd devs unfortunately to review it 😅

sonicaj:
Looking at your diff, that seems good. How do you propose to handle this in iocage for updates ?

grembo:
> Looking at your diff, that seems good. How do you propose to handle this in iocage for updates ?

I thought of adding a function in ioc_common that allows iocage to patch freebsd-update based on the release in question (similar to what I tried first here, but in a more formal way).

skarekrow:
> @araujobsd it is long overdue. We have been looking at regressions from the recent changes ( nat etc ) and i think we are stable enough now. @skarekrow do you think we should perhaps cut one later this week or at the start of the next one ?

I've been pushing for one for some time now ;)

shipit

grembo:
@araujobsd Assuming this will be released soon and you don't change the port before the update to 1.2, please make sure you apply the port skeleton changes from the PR on update (especially the dependency on rcs57).

PR #1031, opened 2019-08-30
NAS-102987 / 11.3 / Add ability to retrieve default properties

sonicaj:
This commit adds ability to retrieve default properties without initializing the iocage dataset.

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -1294,20 +1294,15 @@ class IOCConfiguration(IOCZFS):
                     silent=self.silent)
         return conf
 
-    def check_default_config(self):
-        """This sets up the default configuration for jails."""
-        default_json_location = f'{self.iocroot}/defaults.json'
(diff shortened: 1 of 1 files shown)


skarekrow:
What's the point of this though? Default props are only useful if we're working with jails. So we initialize them then, as we may need to migrate the default props, or jail props if the default props changed.

sonicaj:
If we want to show reasonable defaults of a jail for systems which have integrated iocage, We should be able to do that without initializing the datasets.

skarekrow:
I disagree, this is relying on those datasets to exist to read that configuration. If they're missing, the file doesn't exist to read the default props from.

sonicaj:
If iocage has been initialized, then it's the user/dev's responsibility to expose the `True` defaults as per config. The point is to not initialize the datasets and still be able to expose defaults for the UI

skarekrow (approved):

sonicaj:
Thank you ;)

PR #1033, opened 2019-09-02
Update DF command in faq.rst

badarotti:
Changed the display resource consumption command from
`iocage df [UUID | NAME]`  to `iocage df` 
since that's how it works in iocage v1.1 stable.

$ iocage df test-jail

Usage: iocage df [OPTIONS]
Try "iocage df --help" for help.
Error: Got unexpected extra argument (test-jail)

### DIFF
--- a/doc/source/faq.rst
+++ b/doc/source/faq.rst
@@ -50,7 +50,7 @@ FAQ
     of this documentation for more details.
 
 **Is there a way to display resource consumption?**
-    Yes - :command:`iocage df [UUID | NAME]`
+    Yes - :command:`iocage df`
 
 **Is NAT supported for jails?**
     Yes. NAT is built into FreeBSD. Treat your server as a core


skarekrow:
Thanks! This brings up an interesting query @sonicaj, do we want to bring that functionality back?

badarotti:
> Thanks! This brings up an interesting query @sonicaj, do we want to bring that functionality back?

Would be nice to have both options.

sonicaj:
@badarotti would you like to look into bringing that functionality back ? ; )

badarotti:
> @badarotti would you like to look into bringing that functionality back ? ; )

Sorry for the late reply. I'm afraid I'm not a coder anymore so I can't help there. I'm willing to help with the documentation if you need any!

sonicaj:
@badarotti can you please create a new issue in iocage/issues ? And we can proceed with it considering it a feature. Thank you

sonicaj (approved):

PR #1034, opened 2019-09-04
NAS-103026 / 11.3 / Correctly list admin portal for DHCP/Static IP plugin jails

sonicaj:
This commit introduces changes to ensure that for DHCP/Static ip plugin jails, we don't list the host's ip address in the admin portal. For NAT based jails, it also ensures that the correct nat ports are specified for the plugin jail. However, if the nat_forwards haven't been set by the plugin dev or the end user, we only show the system ip then in that case.

### DIFF
--- a/iocage_lib/ioc_list.py
+++ b/iocage_lib/ioc_list.py
@@ -27,6 +27,7 @@ import netifaces
 import os
 import re
 import subprocess as su
+import urllib.parse
 import uuid as _uuid
 
 import iocage_lib.ioc_common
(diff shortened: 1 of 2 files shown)


iocage_lib/ioc_list.py
```diff
@@ -364,12 +396,36 @@ def list_all(self, jails):
...
+                                # dev hasn't added it to plugin's json yet
+                                nat_port = nat_forwards_dict.get(portal_port)
+                                uri = ''
+                                if portal_uri.scheme:
+                                    uri = f'{portal_uri.scheme}://'
+                                uri += portal_uri.hostname
+                                if nat_port:
+                                    uri += f':{nat_port}'
+                                if portal_uri.path:
+                                    uri += portal_uri.path
```
william-gr: You're missing the other parts, like query.

Why not replace portal_uri.port alone and rebuild the uri with urlunparse?

iocage_lib/ioc_list.py
```diff
@@ -364,12 +396,33 @@ def list_all(self, jails):
...
-                                all_ips
+                        admin_portals = []
+                        for portal in admin_portal.split(','):
+                            if conf.get('nat'):
+                                portal_uri = urllib.parse.urlparse(portal)
+                                portal_port = portal_uri.port or 80
+                                # We do this safely as it's possible
+                                # dev hasn't added it to plugin's json yet
+                                nat_port = nat_forwards_dict.get(portal_port)
+                                netloc = f'{portal_uri._hostinfo[0]}' \
```
william-gr: weird construct, f-string inside a f-string

Perhaps:

```
if nat_port:
    uri = portal_uri._replace(netloc=f'{portal_uri._hostinfo[0]}:{nat_port}').get_url()
else:
    uri = portal
```
Just an idea ; )

william-gr (dismissed):

iocage_lib/ioc_list.py
sonicaj: Agreed ; )

william-gr (approved):
Thanks

PR #1039, opened 2019-09-12
NAS-103095 / 11.3 / Ensure new packagesite is registered by plugins

sonicaj:
1) Makes sure if packagesite is changed in a plugin's manifest, it is correctly reflected in the jail by iocage.
2) Bug fix for ensuring we don't stop jails implicitly when updating plugins or installing packages
3) Bug fix for non-existent function call

### DIFF
--- a/iocage_lib/ioc_plugin.py
+++ b/iocage_lib/ioc_plugin.py
@@ -671,7 +671,9 @@ fingerprint: {fingerprint}
 
     def __fetch_plugin_post_install__(self, conf, _conf, jaildir):
         """Fetches the users artifact and runs the post install"""
-        iocage_lib.ioc_start.IOCStart(self.jail, jaildir, silent=True)
+        status, jid = iocage_lib.ioc_list.IOCList().list_get_jid(self.jail)
+        if not status:
+            iocage_lib.ioc_start.IOCStart(self.jail, jaildir, silent=True)
 
         ip4 = _conf['ip4_addr']
         ip6 = _conf['ip6_addr']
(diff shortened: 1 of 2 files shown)


themylogin (dismissed):

iocage_lib/ioc_plugin.py
```diff
@@ -1159,68 +1159,23 @@ def __update_pkg_remove__(self, jid):
...
-        pkg_repos = plugin_conf["fingerprints"]
 
-        if secure:
-            # Certificate verification
+        try:
+            self.__fetch_plugin_install_packages__(
+                path, plugin_conf, plugin_conf['fingerprints'], [],
+                os.path.join(path, 'root/usr/local/etc/pkg/repos')
+            )
+        except (Exception, RuntimeError):
```
themylogin:
```
>>> issubclass(RuntimeError, Exception)
True
```

iocage_lib/ioc_plugin.py
```diff
@@ -1159,68 +1159,23 @@ def __update_pkg_remove__(self, jid):
...
+            self.__rollback_jail__(name='update')
```
themylogin: Maybe it'll be better to do the logging first?

themylogin (changes requested):

iocage_lib/ioc_plugin.py
```diff
@@ -1159,67 +1159,23 @@ def __update_pkg_remove__(self, jid):
...
-                    {
-                        "level": "EXCEPTION",
-                        "message": msg
-                    },
-                    _callback=self.callback)
-
-        if write:
-            self.json_write(conf)
+            self.__rollback_jail__(name='update')
+            raise e
```
themylogin: Just `raise`

themylogin (changes requested):

themylogin (approved):
Thank you!

PR #1040, opened 2019-09-17
NAS-103127 / 11.3 / Automatically toggle network properties for plugins

sonicaj:
This commit introduces changes where if user supplies a network prop which they would like to use for plugin connectivity, we disable the default network prop supplied by plugin manifest if any.

### DIFF
--- a/iocage_lib/ioc_plugin.py
+++ b/iocage_lib/ioc_plugin.py
@@ -254,8 +254,8 @@ class IOCPlugin(object):
             conf['release'] = re.sub(r"\W\w.", "-", conf['release'])
 
         self.release = conf['release']
-        self.__fetch_plugin_inform__(conf, num, plugins, accept_license)
         props, pkg = self.__fetch_plugin_props__(conf, props, num)
+        self.__fetch_plugin_inform__(conf, num, plugins, accept_license)
         location = f"{self.iocroot}/jails/{self.jail}"
 
         try:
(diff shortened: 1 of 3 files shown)


skarekrow:
This shouldn't be needed, as plugins shouldn't have network props defined. The few that do should require them (like vnet), so this would cause regressions with those.

sonicaj:
For having all plugins default to using `nat`. We are specifying `nat`/`dhcp` properties in the plugin manifests. Could you please elaborate on the regressions this can have ?

skarekrow:
We shouldn’t be doing that. We should be defaulting to nat or only overriding nat here. Vnet is the bigger worry. Some plugins (can’t recall off the top of my head) require it to function 100%. Plex is one, as the multicasts and unicasts are required.

sonicaj:
I am not sure i follow, we have `vnet` enabled, what's the issue ?
Are you saying that some plugins won't function well with `nat` ?

skarekrow:
I’m saying some won’t function well *without* vnet. 

We should only be overriding nat here. Nothing else network wise. And we should not be putting random network props (like dhcp) in manifests as well.

william-gr:
They are default props, users can override them

sonicaj:
Okay so we are not disabling `vnet` anywhere ?

This PR only introduces changes to `nat`/`dhcp`/`ip4_addr` and `ip6_addr` properties

iocage_lib/ioc_plugin.py
```diff
@@ -469,11 +486,31 @@ def __fetch_plugin_props__(self, conf, props, num):
...
+        ) and (not iocage_lib.ioc_common.lowercase_set(
+                iocage_lib.ioc_common.construct_truthy('dhcp')
+        ) & iocage_lib.ioc_common.lowercase_set(
+            create_props) and not iocage_lib.ioc_common.lowercase_set(
+            iocage_lib.ioc_common.construct_truthy('ip_hostname')
+        ) & iocage_lib.ioc_common.lowercase_set(
+            create_props) and not iocage_lib.ioc_common.lowercase_set(
+            iocage_lib.ioc_common.construct_truthy('nat')
+        ) & iocage_lib.ioc_common.lowercase_set(create_props)):
+            iocage_lib.ioc_common.logit(
```
skarekrow, line 499: I regret writing blocks like this, perhaps we should make this a common function so it reads much nicer?

skarekrow (dismissed):
Meh, you're right. I missed a detail on this.

iocage_lib/ioc_plugin.py
sonicaj: Roger that :p

skarekrow (approved):
So much nicer!

PR #1046, opened 2019-09-24
NAS-102932 / 11.3 / Remove py-libzfs from iocage

sonicaj:
This PR aims to remove py-libzfs from iocage and introduce a new api to handle zfs related tasks.


iocage_lib/ioc_plugin.py
```diff
@@ -1429,9 +1427,6 @@ def __check_manifest__(self, plugin_conf):
                 },
                 _callback=self.callback)
 
-        if write:
-            self.json_write(plugin_conf)
```
william-gr, line 1433: Wondering why this changed?

iocage_lib/ioc_json.py
```diff
@@ -678,31 +681,29 @@ def get_version():
...
-                    match += 1
-                elif prop_comment == "iocage":
-                    _dataset = pool
-                    match += 1
+            matches = []
+            for pool in PoolListableResource():
+                ds = Dataset(pool.name)
+                if pool.active:
+                    matches.append(pool)
+                elif ds.properties['comment'] == 'iocage':
```
william-gr: Isnt comment a user property? Doesnt this need special handling? e.g. org.iocage:comment ? or does it already do that

iocage_lib/ioc_plugin.py
sonicaj: I don't know why it was there in the first place :p ioc plugin has no function with that signature

iocage_lib/ioc_json.py
sonicaj: It indeed is but keeping the old implementation, it has not followed that convention. it just is `comment`

iocage_lib/dataset.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @property
+    def dataset(self):
+        return Dataset(self.resource_name.split('@', 1)[0])
+
+    def create_snapshot(self, options=None):
+        return create_snapshot(self.resource_name, options)
+
+    @property
+    def exists(self):
```
william-gr, line 104: This feels a bit weird to be a property since it requires non-trivial computation to be evaluated.

Same for dataset.

iocage_lib/ioc_fetch.py
```diff
@@ -251,16 +253,16 @@ def fetch_release(self, _list=False):
...
 
             for f in self.files:
                 if not os.path.isfile(f):
 
-                    _dataset = self.zfs.get_dataset(pool_dataset)
-
-                    _dataset.umount()
-                    _dataset.delete()
+                    ds = Dataset(pool_dataset)
+                    ds.umount()
```
william-gr: I dont think you need umount here, zfs destroy does it. umount was required in py-libzfs

iocage_lib/ioc_json.py
william-gr: I dont follow. `comment` is not a valid zfs property, how can you expect it to exist?

william-gr (dismissed):

iocage_lib/ioc_json.py
sonicaj: I don't. I am just keeping the old behavior. Perhaps @skarekrow can elaborate ?

skarekrow: It was used for legacy iocage installations. The property is a *zpool* property.

sonicaj: Well in that case why are we checking it for pool root dataset ?

skarekrow: I don't remember the reason why, it was years ago :P

william-gr: Lets remove it then, it has been broken for years :D

william-gr (approved):

themylogin:
Good job, this is giant and after several reviews I could not find anything notable

PR #1057, opened 2019-10-10
NAS-103410 / 11.3 / Retrieve more data with iocage list

sonicaj:
This PR introduces changes which adds the ability for the API to return more data related to plugins. It helps reduce retrieval times when we try to find source of a plugin.

### DIFF
--- a/iocage_lib/ioc_list.py
+++ b/iocage_lib/ioc_list.py
@@ -46,9 +46,10 @@ class IOCList(object):
         JID UID BOOT STATE TYPE IP4 RELEASE
     """
 
-    def __init__(self, lst_type="all", hdr=True, full=False, _sort=None,
-                 silent=False, callback=None,
-                 plugin=False, quick=False):
+    def __init__(
+        self, lst_type='all', hdr=True, full=False, _sort=None, silent=False,
+        callback=None, plugin=False, quick=False, **kwargs
+    ):
         self.list_type = lst_type
         self.header = hdr
         self.full = full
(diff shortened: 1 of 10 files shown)


iocage_lib/ioc_list.py
```diff
@@ -455,6 +461,25 @@ def list_all(self, jails):
                 jail_list.append([jid, uuid, boot, state, jail_type,
                                   full_release, full_ip4, ip6, template,
                                   admin_portal, doc_url])
+                if self.plugin_data:
```
skarekrow, line 464: We should probably want the remote flag set for this to occur, as list is expecting all resources to be listed are local and quick. This could have a long delay if any network disruptions are occurring or anything else is funky network wise.

sonicaj: Are you referring to `iocage list` ?

skarekrow: Yeah if this is ever exposed to cli behavior.

sonicaj: I don't think that it will have much if any use there. Also we could be perhaps increasing more props here, so the formatting is not going to work at all. I doubt it will even with just 3 more props. Plus the repo uri's are quite big ; ) I think we are good for now. However wrt the PR, i am thinking we should do the loop asynchronously, will look more into that tomorrow. Marking as WIP :p

william-gr (dismissed):

william-gr (approved):

PR #1063, opened 2019-10-14
Automatically set server && root_dir when fetching EoL releases

### LINKED ISSUE #1059: Fetch old releases from ftp-archive.FreeBSD.org
Jails are an awesome way to explore old FreeBSD releases, and iocage is the easiest way to create a jail.  But the current implementation fetches releases from https://download.freebsd.org, which only serves currently supported releases.  As an enhancement, it would be awesome if iocage could fetch historic releases from ftp-archive.FreeBSD.org .  This recently came up at work, when I needed to install a 9.3 jail.
```
> iocage --version
Version	1.2 RC
> sudo iocage fetch -r 9.3-RELEASE
Password:
9.3-RELEASE was not found!
```

asomers:
When fetching an EoL release, iocage will now automatically set the
server and root dir appropriately.  This allows fetching releases as old
as 9.0-RELEASE.  Older releases are still unfetchable because the
release files used a different format.

Fixes #1059

- [x] Explain the feature


iocage_lib/ioc_fetch.py
```diff
@@ -300,12 +301,22 @@ def fetch_http_release(self, eol, _list=False):
...
         if self.hardened:
-            if self.server == "download.freebsd.org":
+            if self.server is None:
                 self.server = "http://jenkins.hardenedbsd.org"
                 rdir = "builds"
 
-        if self.root_dir is None:
-            self.root_dir = f"ftp/releases/{self.arch}"
+        md = re.match("([0-9]+)[.]([0-9]+)-RELEASE", self.release)
+        if md and (int(md[1]) < 11 or (int(md[1]) == 11 and int(md[2]) < 2)):
```
jsegaert, line 309: This will require maintenance in perpetuity.  
Instead of hard-coding 11.2, maybe consider something more dynamic as in function `ioc_common.parse_latest_release()` that gets the latest release.

asomers: The problem with that approach is that it would necessitate an additional round trip to the internet.  I don't think we want to do that.

skarekrow:

asomers:
No @skarekrow I don't think I have enough time to work on this.  I'm going to abandon the PR, even though I still dream about someday doing `iocage create -r 1.0-RELEASE`.

PR #1067, opened 2019-10-21
NAS-103568 / 11.3 / Inherit mount point from parent

sonicaj:
This commit fixes an issue where we set the mount point as that of the parent if iocage's mount point didn't start with it. But if altroots are involved, they are added automatically to the mount point resulting in a misconfigured mount point.


skarekrow (approved):

sonicaj:
backport

sonicaj:
backport

PR #1077, opened 2019-10-24
NAS-103568 / 11.3 / Bug fixes for fstab

sonicaj:
This PR fixes following bugs:

1) Iocage did not judge destinations correctly when attempting to correct malformed fstab
2) Iocage wrote python list to fstab if fstab file's contents were changed during validation.
3) Add `/mnt/mnt` to list of malformed mount points for fstab correction.

### DIFF
--- a/iocage_lib/ioc_fstab.py
+++ b/iocage_lib/ioc_fstab.py
@@ -439,7 +439,8 @@ class IOCFstab(object):
                             _callback=self.callback,
                             silent=self.silent)
                     else:
-                        fstab.write(f'{line}\n')
+                        is_list = isinstance(line, list)
+                        fstab.write(f'{line[1] if is_list else line}\n')
 
             if not matched:
                 iocage_lib.ioc_common.logit({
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_fstab.py
```diff
@@ -228,8 +231,11 @@ def __validate_fstab__(self, fstab, mode='single', actions=None):
 
             if not source.is_dir():
                 if fstype == 'nullfs':
-                    if pathlib.Path('/mnt/iocage') in source.parents or \
-                            pathlib.Path('/iocage') in source.parents:
+                    if any(
+                        pathlib.Path(i) in source.parents for i in (
```
skarekrow, line 235: What about altroots that aren't any of these. Or different mountpoints?

sonicaj: All of these scenarios are cases where we are correcting past iocage behavior, do you have any other examples for the scenarios you mention which we explicitly created ?

skarekrow: No just me not reading the diff closely enough 😂

skarekrow (approved):

iocage_lib/ioc_fstab.py
sonicaj: Hahaha :D No worries, happens to all of us ;)

william-gr (approved):

PR #1079, opened 2019-10-25
Update .Dd date

dlangille:
Let's see if this works properly.  new iocage.8 and iocage.8.gz


skarekrow (approved):
Thanks Dan and @asomers for catching this!

PR #1082, opened 2019-10-27
Initial support for anchors in pf nat backend

### LINKED ISSUE #1081: nat with pf backend overrides host config
When running iocage with `nat_backend="pf"` the host systems pf rules are ignored. The nat functionality just loads it's temporary pf.conf and overrides any previously applied configuration.

It is possibly the same behavior with ipfw (did not check).

```
root@host:~ # FreeBSD host 12.0-RELEASE-p10 FreeBSD 12.0-RELEASE-p10 GENERIC  amd64
root@host:~ # iocage --version
Version 1.2
root@host:~ # pkg version | grep py36-iocage
py36-iocage-1.2_2
```
```
root@host:~ # pfctl -sr
scrub in on bridge0 all fragment reassemble
[...] (lengthy list of pf-Rules)
root@host:~ # pfctl -sn
root@host:~ # iocage get nat testjail
1
root@host:~ # iocage get nat_backend testjail
pf
root@host:~ # iocage start testjail
testjail: nat requires nat_interface, using bridge0
* Starting testjail
  + Started OK
  + Using devfs_ruleset: 5
  + Configuring VNET OK
  + Using IP options: vnet
  + Starting services OK
  + Executing poststart OK
root@host:~ # pfctl -sr
root@host:~ # pfctl -sn 
nat on bridge0 inet from 172.16.0.0/24 to any -> (bridge0:0) static-port
```

I definitely don't expect iocage to override my pf configuration. At least this behavior needs to be documented.
I understand that it is not trivial to apply pf rules for jails into the correct line of custom rule sets, but a possible fix for this would be to search the current pf rules for specific anchors (like `iocage_nat` and `iocage_rdr`) and if they are found, apply the rules within these anchors. This way one can choose to create those anchors at the right place in it's own pf.conf in order to apply iocage pf rules in the correct place. In case the anchors are not found iocage may still fall back to the current behavior (and maybe issue a warning).
**If the general idea is acceptable for you I may create a fix and send in a PR.**

 a stable release.

Since the current masters code is the same, the expected result is the same.

haemka:
Fixes #1081 

As proposed in #1081 this change inserts iocage nat rules into specific anchors if these exist.


haemka:
So the tests seem to fail with a module error for sqlite3 initiated by pytest. I'm pretty sure this is not introduced by my changes. Are your tests working correctly?

skarekrow:

PR #1084, opened 2019-10-28
NAS-102724 / 11.3 / Retrieve maintainer when retrieving plugin data

sonicaj:


william-gr (commented):
Not sure we want to default maintainer to iX if there isnt any?

Should we say there isnt a maintainer instead?

william-gr (approved):

PR #1091, opened 2019-11-06
Correct typo in error message for iocage set

gzachv:
Fix typo in error message displayed when not providing a value for a
setting during an 'iocage set' operation.

Previously, this appeared as the following:
```
# iocage set -P foo test
foo is is missing a value!
#
```


skarekrow (approved):
Thanks!

PR #1092, opened 2019-11-07
NAS-103827 / 11.3 / Allow users to use higher minor version number jails

sonicaj:
This PR introduces changes which allow user to use higher minor version number jails. This is allowed by FreeBSD and is okay to do so.

### DIFF
--- a/iocage_cli/create.py
+++ b/iocage_cli/create.py
@@ -164,7 +164,7 @@ def cli(
 
     if release:
         try:
-            ioc_common.check_release_newer(release)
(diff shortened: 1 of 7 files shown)


sonicaj:
@skarekrow can you review this one please ?

skarekrow (approved):
I agree, are there situations we care about testing that it's not major only? Perhaps we set that as default and then adjust those few situations to supply False.

sonicaj:
@skarekrow i think it's better safe to be sorry - i have gone through most of the usages and explicitly specified this.

PR #1093, opened 2019-11-07
ioc_start: More thorough check of requested IP against current IPs

### LINKED ISSUE #1018: ioc_start: IPAddr comparison fails to take /netmask into account
In the case where an IP address is given without an interface (Shared IP), it is valid to pass a netmask, eg. `192.168.1.1/24` according to `jail(8)`.

In this case, iocage networking will break as a comparison when starting the jails ends up being `192.168.1.1/24 in ['192.168.1.1', etc]`. iocage will then proceed to configure the IP address on what it determined to be the default interface, even though the address already exists elsewhere.

https://github.com/iocage/iocage/blob/d298ad4a2d554a756985e656e74eebe16e013f16/iocage_lib/ioc_start.py#L938

phyber:
Fixes #1018

This PR improves the `current_ips` check in `ioc_start`, in order to take into account IP netmasks.
Previously, the `if ip in current_ips` check could fail to recognise an IP address already existed on an interface if `ip` was `127.0.0.1/24` and `current_ips` was `['127.0.0.1']`.

It is valid for `ip` to contain a netmask, as documented in `jail(8)`, so we must try harder when checking if the address is already on an interface or not.

This resolves an issues such as the following:

  - A host has two or more interfaces configured
  - One interface is external (`ext0`)
  - One interface is internal (`int0`)
  - `ext0` is public, internet facing, and has the default route
  - `int0` is private, internal, has some IPs on it for jails to share with the host

If we were to bring up a jail while requesting an IP configured on `int0` such as `192.168.1.1/24`, `iocage` would previously fail to recognise that the address was configured on` int0` and proceed to add it as an alias to `ext0`.

- [x] Explain the feature


skarekrow:

phyber:
Reopened by request! Or not. I thought replying did it. Hmm.

phyber:
I'll take a look at rebasing this and opening a new PR.

PR #1094, opened 2019-11-07
NAS-101246 / 11.3 / Ensue accept_rtadv is respected and jail gets an ip6

sonicaj:


skarekrow (approved):

Oriann:
Tested in latest build from master tree, but still does not work.
VNET is broken and so jail wont even start with vnet=on

sonicaj:
@Oriann can you please make a new issue describing what your configuration is, what the expected result is and what happens please ? If you are using FreeNAS, it would be better to create a ticket in jira and attach a debug. Thank you

PR #1095, opened 2019-11-08
NAS-106037 / 12.0 / Make MTU of VNet interfaces configurable

nathanjrobertson:
This patch adds five new configuration parameters, with the default behaviour being identical to the pre-existing behaviour (copy the MTU value from the bridge, or 1500 if there is no existing bridge). We add:

* vnet[0-3]_mtu - default to "auto" (ie. existing behaviour). If set to a value, it will take top precedence in creating the device (ie. it will ignore the value the bridge has)
* vnet_default_mtu - replaces the hardcoded "1500" value in the code with a configurable variable (but the variable defaults to "1500").

The use case we have for this where a bridge (say bridge0) has an IP address, and we bring it up on boot with that address (but no bridged NIC attached). if_bridge(4) says that you can't set the MTU on bridge0 with ifconfig - it is configured from the first attached device to the bridge. In our case, the VNet device ends up being the first attached, and this is hardcoded to 1500 (the VPS provider the machine runs on requires it to be 1450). Using one of the two above options (it seemed sensible to provide either a global or per-interface configuration) covers this situation.

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -895,8 +895,8 @@ class IOCConfiguration:
         for x in range(0, 4):
             if not conf.get(f"vnet{x}_mtu"):
                 conf[f"vnet{x}_mtu"] = 'auto'
-        if not conf.get(f"vnet_default_mtu"):
-            conf[f"vnet_default_mtu"] = '1500'
+        if not conf.get("vnet_default_mtu"):
+            conf["vnet_default_mtu"] = '1500'
 
         if not default:
             conf.update(jail_conf)
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -863,7 +863,7 @@ class IOCConfiguration:
                 conf[option] = 'auto'
 
         # Version 27 keys
-        for x in range(0,4):
+        for x in range(0, 4):
             if not conf.get(f"vnet{x}_mtu"):
                 conf[f"vnet{x}_mtu"] = 'auto'
         if not conf.get(f"vnet_default_mtu"):
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -863,11 +863,11 @@ class IOCConfiguration:
         for option in ('defaultrouter', 'defaultrouter6'):
             if conf.get(option) == 'none':
                 conf[option] = 'auto'
-        
+
         # Version 27 key
         if not conf.get('min_dyn_devfs_ruleset'):
             conf['min_dyn_devfs_ruleset'] = '1000'
-        
+
         # Version 28 keys
         for x in range(0, 4):
             if not conf.get(f"vnet{x}_mtu"):
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -433,7 +433,7 @@ class IOCConfiguration:
     @staticmethod
     def get_version():
         """Sets the iocage configuration version."""
-        version = '26'
+        version = '27'
 
         return version
 
@@ -862,6 +862,13 @@ class IOCConfiguration:
             if conf.get(option) == 'none':
                 conf[option] = 'auto'
 
+        # Version 27 keys
+        for x in range(0,4):
+            if not conf.get(f"vnet{x}_mtu"):
+                conf[f"vnet{x}_mtu"] = 'auto'
+        if not conf.get(f"vnet_default_mtu"):
+            conf[f"vnet_default_mtu"] = '1500'
+
         if not default:
             conf.update(jail_conf)
 
@@ -1190,6 +1197,11 @@ class IOCConfiguration:
             'nat_forwards': 'none',
             'plugin_name': 'none',
             'plugin_repository': 'none',
+            'vnet0_mtu': 'auto',
+            'vnet1_mtu': 'auto',
+            'vnet2_mtu': 'auto',
+            'vnet3_mtu': 'auto',
+            'vnet_default_mtu': '1500',
         }
 
     def check_default_config(self):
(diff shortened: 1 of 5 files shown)


skarekrow (dismissed):
LGTM, @sonicaj?

nathanjrobertson:
@skarekrow Tests now pass. Could you please review again? Thanks.

skarekrow (dismissed):
Still LGTM

nathanjrobertson:
@sonicaj Any feedback? I'd very much like this patch to make the next iocage release, as we're using it in production currently.

nathanjrobertson:
@skarekrow I had to merge a conflict in iocage_json.py, as there was another configuration option added which took the version number I was using. No other changes, just a merge. Could you please review again and approve.

skarekrow (dismissed):
Still LGTM, @sonicaj?

nathanjrobertson:
@skarekrow Updated to pass travis tests (whitespace warnings only).

skarekrow (dismissed):

nathanjrobertson:
@sonicaj Could you please take a look at this and consider merging it? I realize releases of iocage are infrequent, and that merging won't make it into FreeBSD's ports collection for a fair while, but every time there's a package update it reminds me that having to patch manually all the time is getting a bit old. And I really don't want this to miss the next release, because I've got production servers which rely on this patch. Thanks.

iocage_lib/ioc_json.py
```diff
@@ -868,6 +868,13 @@ def check_config(self, conf, default=False):
         if not conf.get('min_dyn_devfs_ruleset'):
             conf['min_dyn_devfs_ruleset'] = '1000'
 
+        # Version 28 keys
+        for x in range(0, 4):
+            if not conf.get(f"vnet{x}_mtu"):
+                conf[f"vnet{x}_mtu"] = 'auto'
+        if not conf.get(f"vnet_default_mtu"):
```
sonicaj: This does not have to be a `f-string`

iocage_lib/ioc_start.py
```diff
@@ -1500,7 +1502,9 @@ def find_bridge_mtu(self, bridge):
 
         memberif = self.get_bridge_members(bridge)
         if not memberif:
-            return '1500'
+            if self.unit_test:
+                return "1500"
```
sonicaj: Can you please elaborate why we have this ? Wouldn't the default be `1500` and if no, don't you think this should be handled in the unit test rather then here ?

sonicaj (commented):
Thank you @nathanjrobertson, a few minor nits

iocage_lib/ioc_json.py
nathanjrobertson: Yep, fair point. Got a bit keen on the copy/paste. I've pushed a patch for this one to only use f-strings where there is variable substitution.

iocage_lib/ioc_start.py
nathanjrobertson: I'll take a look at this further. The patch changes the default to be set by a variable, but that variable doesn't get set in unit_test mode. There were other places in the code where this stubbing out for test mode happened, so it seemed to be the way it was done. It didn't seem right to me, but seemed to be a convention. Anyway, leave it with me and I'll come up with something more elegant, because I didn't like that when I did it, so I'm glad there is a better way of handling unit_test mode.

In the mean time, it seems that Cirrus CI is failing on installing packages (ie. before our tests start).

nathanjrobertson: @sonicaj Ok, I've reproduced the error when I remove that "if self.unit_test: return "1500". It needs that because otherwise we get

> AttributeError: 'IOCStart' object has no attribute 'get'

in when running test mode (tests/unit_tests/1000_lib_start_test.py:55). The patch I'm looing to have merged makes the default MTU configurable, and that gets done with an IOCStart.get() call. So, I can see three options:

1. Leave it as is, with a hardcoded default that only services the unit test. This is done 30 lines above in the same function (dealing with DHCP being unavailable in unit test mode).
2. Remove the test. The test is hardcoded to test for "1500", and hardcoding the return value in #1 has very limited value in testing.
3. Remove the feature of having a default MTU other than 1500. The value of the default used to be hardcoded to 1500 (in fact, the whole thing was hardcoded to 1500 without this patch), so having a hardcoded default as long as you can configure per-interface is still useful.

I'd personally recommend solution #2. Remove what is a test which has limited value, particularly given the default can be changed. Are you ok with that being the resolution?

nathanjrobertson: Actually, please don't do #3. Having a default MTU setting is actually more useful than having per-interface. Chances are, in a given environment, it'll be the same for all interfaces, so you'll end up setting each and every interface to whatever the value you need is. Possible, but more painful.

nathanjrobertson: @sonicaj I've just committed an update which I'd like reviewed. I've provided a dummy IOCStart.get() for that test. So, the test remains, and the test does something useful. Could you please review again?

I'm just waiting for Cirrus-CI to finish executing. I needed to update it to an 11.3-STABLE snapshot in order to get it to execute. The version of the packages in the FreeBSD package repo are for 11.3, and don't install on 11.2 any more, it seems.

nathanjrobertson: Ok. Test is fixed so that it does test the functionality, and the CirrusCI tests pass. I'll resolve this thread.

nathanjrobertson:
@sonicaj I've made the changes requested, and have the CirrusCI tests running as well (they were broken across the project - the breakage wasn't related specifically to the changes I made). In fixing that, I had to update us to 11.3-RELEASE in the jail and a snapshot of 11.3-STABLE to get the tests running again (apparently CirrusCI uses gCloud, which doesn't currently work with 11.3-RELEASE, but does with -STABLE snapshots). The PR should be ready to consider for merging now.

nathanjrobertson:
@skarekrow @sonicaj Could somebody please take another look at this and reconsider for merging?

sonicaj (approved):
Thank you @nathanjrobertson this looks good to me.

PR #1097, opened 2019-11-08
Correct typo in error message for iocage set

gzachv:
Fix typo in error message displayed when not providing a value for a
setting during an 'iocage set' operation.


sonicaj (approved):
Thank you @gzachv

PR #1106, opened 2019-12-05
NAS-104777 / 12.0 / Change devfs ruleset handling so that configured rulesets != 4 are

grembo:
cloned/copied into dynamic ones.  This makes devfs rule handling more
symmetrical to what is done when the default ruleset 4 is configured (which
in fact never applies devfs ruleset 4, but creates an iocage specific
dynamic ruleset instead - which can be quite confusing).

As a result, this addresses the problem of non-dynamic rulesets being
removed on `iocage stop` raised in https://github.com/iocage/iocage/issues/952.

This also makes iocage fail to start a jail if the configured devfs ruleset
is not available - beforehand it would start with a default ruleset in this
case, which can have severe unwanted side effects.

Finally, this sets the minimum dynamically assigned devfs rule id to 1000 to
reserve the lower ids for static configuration by the admin in devfs.rules
(this is mostly for convenience).

- [x] Explain the feature

### DIFF
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -49,8 +49,6 @@ INTERACTIVE = False
 # 4 is a magic number for default and doesn't refer
 # to the actual ruleset 4 in devfs.rules(!)
 IOCAGE_DEVFS_RULESET = 4
-# Leave lower devfs_ruleset ids to sysadmin
-MIN_DYNAMIC_DEVFS_RULESET = 1000
 
 
 def callback(_log, callback_exception):
@@ -755,7 +753,7 @@ def generate_devfs_ruleset(conf, paths=None, includes=None, callback=None,
     )
     ruleset_list = [int(i) for i in devfs_rulesets.stdout.splitlines()]
 
-    ruleset = MIN_DYNAMIC_DEVFS_RULESET
+    ruleset = int(conf["min_dyn_devfs_ruleset"])
     while ruleset in ruleset_list:
         ruleset += 1
     ruleset = str(ruleset)
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -740,7 +740,7 @@ def generate_devfs_ruleset(conf, paths=None, includes=None, callback=None,
     Will add a per jail devfs ruleset with the specified rules,
     specifying defaults that equal devfs_ruleset 4
     """
-    ruleset = conf['devfs_ruleset']
+    configured_ruleset = conf['devfs_ruleset']
     devfs_includes = []
     devfs_rulesets = su.run(
         ['devfs', 'rule', 'showsets'],
@@ -748,22 +748,37 @@ def generate_devfs_ruleset(conf, paths=None, includes=None, callback=None,
     )
     ruleset_list = [int(i) for i in devfs_rulesets.stdout.splitlines()]
 
-    if ruleset != '4':
-        if int(ruleset) in ruleset_list:
-            return str(ruleset)
+    # 4 is a magic number for default and doesn't refer to
+    # the actual ruleset 4 in devfs.rules(!)
+    default_ruleset = 4
+    min_dynamic_ruleset = 1000  # leave lower numbers to sysadmin
 
-        logit({
-            "level": "INFO",
-            "message": f'* Ruleset {ruleset} does not exist, using defaults'
-        },
-            _callback=callback,
-            silent=silent)
-
-    ruleset = 5  # 0-4 is always reserved
+    ruleset = min_dynamic_ruleset
     while ruleset in ruleset_list:
         ruleset += 1
     ruleset = str(ruleset)
 
+    # Custom devfs_ruleset configured, clone to dynamic ruleset
+    if int(configured_ruleset) != default_ruleset:
+        if int(configured_ruleset) not in ruleset_list:
+            logit(
+                {
+                    "level": "EXCEPTION",
+                    "message": f"devfs_ruleset {ruleset} doesn't exist"
+                },
+                _callback=callback,
+                silent=silent)
+        rules = su.run(
+            ['devfs', 'rule', '-s', configured_ruleset, 'show'],
+            stdout=su.PIPE, universal_newlines=True
+        )
+        for rule in rules.stdout.splitlines():
+            su.run(['devfs', 'rule', '-s', ruleset, 'add'] +
+                   rule.split(' ')[1:], stdout=su.PIPE)
+
+        return ruleset
+
+    # Create default ruleset
     devfs_dict = dict((dev, None) for dev in (
         'hide', 'null', 'zero', 'crypto', 'random', 'urandom', 'ptyp*',
         'ptyq*', 'ptyr*', 'ptys*', 'ptyP*', 'ptyQ*', 'ptyR*', 'ptyS*', 'ptyl*',
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -44,6 +44,11 @@ import iocage_lib.ioc_exceptions
 import iocage_lib.ioc_exec
 
 INTERACTIVE = False
+# 4 is a magic number for default and doesn't refer
+# to the actual ruleset 4 in devfs.rules(!)
+IOCAGE_DEVFS_RULESET = 4
+# Leave lower devfs_ruleset ids to sysadmin
+MIN_DYNAMIC_DEVFS_RULESET = 1000
 
 
 def callback(_log, callback_exception):
@@ -748,26 +753,15 @@ def generate_devfs_ruleset(conf, paths=None, includes=None, callback=None,
     )
     ruleset_list = [int(i) for i in devfs_rulesets.stdout.splitlines()]
 
-    # 4 is a magic number for default and doesn't refer to
-    # the actual ruleset 4 in devfs.rules(!)
-    default_ruleset = 4
-    min_dynamic_ruleset = 1000  # leave lower numbers to sysadmin
-
-    ruleset = min_dynamic_ruleset
+    ruleset = MIN_DYNAMIC_DEVFS_RULESET
     while ruleset in ruleset_list:
         ruleset += 1
     ruleset = str(ruleset)
 
     # Custom devfs_ruleset configured, clone to dynamic ruleset
-    if int(configured_ruleset) != default_ruleset:
+    if int(configured_ruleset) != IOCAGE_DEVFS_RULESET:
         if int(configured_ruleset) not in ruleset_list:
-            logit(
-                {
-                    "level": "EXCEPTION",
-                    "message": f"devfs_ruleset {ruleset} doesn't exist"
-                },
-                _callback=callback,
-                silent=silent)
+            return (True, configured_ruleset, '0')
         rules = su.run(
             ['devfs', 'rule', '-s', configured_ruleset, 'show'],
             stdout=su.PIPE, universal_newlines=True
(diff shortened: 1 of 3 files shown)


grembo:
@sonicaj Do you think there is a chance to look into this any time soon?

sonicaj:
@skarekrow would you have some time to have a peek at this please ?
@grembo I'll try to review this coming weekend. Apologies for the delay

skarekrow:
I don’t have much free time unfortunately. Possibly next weekend, but no promises :(

grembo:
To give some more background: The previous protecting behaviour was removed in https://github.com/iocage/iocage/commit/19cc401d9ffc8b7635a3d3214e1ac5a07fe05428#diff-2be5e949a8b1c257fbb3dbc48a059c90 as a consequence of another change. Please note that this is not only an inconvenience, but in some scenarios could also cause security problems (in cases where a strict ruleset is removed and replaced by the more permissive default one - or a firewall device isn't made available, leaving a vnet jail not firewalled - all triggered by an innocent iocage restart jailname).

grembo:
Is there anything I can do to help?

iocage_lib/ioc_common.py
```diff
@@ -740,30 +740,45 @@ def generate_devfs_ruleset(conf, paths=None, includes=None, callback=None,
...
         ruleset += 1
     ruleset = str(ruleset)
 
+    # Custom devfs_ruleset configured, clone to dynamic ruleset
+    if int(configured_ruleset) != default_ruleset:
+        if int(configured_ruleset) not in ruleset_list:
+            logit(
+                {
+                    "level": "EXCEPTION",
+                    "message": f"devfs_ruleset {ruleset} doesn't exist"
```
sonicaj: @grembo can we please catch the exceptions which this would raise in `ioc_start.py` and raise an appropriate error message saying that `Failed to start {jail_name}: the current one we have right now here`. Thoughts ?

sonicaj (commented):
Thank you @grembo for the contribution. This looks good to me and good catch ;)
I just have a minor nit

iocage_lib/ioc_common.py
grembo: Please see comment on not-outdated commit...

grembo:
Hi @sonicaj,

Thanks for looking into this. I changed the code now, so it shows a proper error message in case a devfs_ruleset isn't configured.

While doing so I found a couple of more issues that I addressed, please see the commit message of the new patch. This makes the patch a bit bigger, but also *a lot* more useful. 

Let me know if you have any questions.

See below on what startup messages look like now.

devfs_ruleset non-existent:
```
# iocage start myjail
* Starting myjail
myjail devfs_ruleset 777 does not exist! - Not starting jail
```

devfs_ruleset existent:
```
# iocage start myjail
* Starting myjail
  + Started OK
  + Cloning devfs_ruleset 501
  + Using devfs_ruleset: 1006
  + Using IP options: ip4.addr=lo1|10.0.111.21/32...
  + Starting services OK
  + Executing poststart OK
```

devfs_ruleset default/unconfigured:
```
# iocage start myjail
* Starting myjail
  + Started OK
  + Generating dynamic devfs_ruleset
  + Using devfs_ruleset: 1006
  + Using IP options: ip4.addr=lo1|10.0.111.21/32...
  + Starting services OK
  + Executing poststart OK
```

iocage_lib/ioc_start.py
```diff
@@ -621,6 +646,22 @@ def __start_jail__(self):
                 _callback=self.callback,
                 silent=self.silent)
 
+        if manual_devfs_config:
```
sonicaj: These statements are inaccurate and if included should be included at the place where what they say is happening actually happens ( which is above ).

iocage_lib/ioc_start.py
```diff
@@ -512,17 +503,51 @@ def __start_jail__(self):
...
+                "message": f"{self.uuid} devfs_ruleset"
+                           f" {configured_devfs_ruleset} does not exist!"
+                           " - Not starting jail"
+            },
+                _callback=self.callback,
+                silent=self.silent)
+            return
+
+        # Manually configured devfs_ruleset doesn't support all iocage features
+        if manual_devfs_config:
```
sonicaj, line 527: I am not sure if we should increase verbosity to this level for something which the user deliberately configured. Thoughts @skarekrow ? In my opinion we shouldn't add these

sonicaj (commented):
Hi @grembo, some minor nits.

iocage_lib/ioc_start.py
skarekrow: They may have done so erroneously, so being aware of what the ramifications are should help to make better decisions so I’m in favor of this.

iocage_lib/ioc_common.py
```diff
@@ -46,6 +46,11 @@
 from iocage_lib.dataset import Dataset
 
 INTERACTIVE = False
+# 4 is a magic number for default and doesn't refer
+# to the actual ruleset 4 in devfs.rules(!)
+IOCAGE_DEVFS_RULESET = 4
+# Leave lower devfs_ruleset ids to sysadmin
+MIN_DYNAMIC_DEVFS_RULESET = 1000
```
skarekrow: I think this would be better as a property for iocage that way the user can configure this the way they do everything else.

skarekrow (changes requested):

iocage_lib/ioc_common.py
grembo: That would be a global configuration though. Where would you do that?

skarekrow: In the same place as normal properties, and we would just set the default as 1000. Allowing each jail to have their own minimum floor.

iocage_lib/ioc_start.py
grembo: They absolutely should be included, otherwise it's super confusing. I noticed this myself while testing earlier after not looking at this for quite a while - only showing the dynamic ruleset id is confusing, as you don't know where it was copied from. It's the same for the default one (4) - this also confusing while debugging originally, like "what is this ruleset?".

I also wanted to make sure people understand that ruleset 4 isn't at all related to devfs_ruleset 4 on the system (they happen to server the same purpose).

I also didn't want to change the original start output (people might rely on it already somehow), otherwise I would've put it in parentheses behind it.

What about something like this?

```
 + Configured devfs to clone from ruleset {configured_devfs_ruleset}
 + Using devfs_ruleset: {devfs_ruleset}
```

and

```
 + Configured devfs to generated default ruleset
 + Using devfs_ruleset: {devfs_ruleset}
```

If changing the output is okay, we could also do

```
 + Using devfs_ruleset: {devfs_ruleset} (cloned from {configured_devfs_ruleset})
```

and

```
 + Using devfs_ruleset: {devfs_ruleset} (iocage generated)
```

skarekrow: I prefer changing the output, relying on userland  output is never a good idea so users should expect things to change. It’s much cleaner than adding yet another log line and this information does need to be present.

grembo: Done

iocage_lib/ioc_common.py
grembo: Done

iocage_lib/ioc_start.py
grembo: Half of that check was in there before [DHCP] (I just moved it to the same block to take advantage of having only one check if a devfs ruleset was manually configured). I also think that having that kind of check in there could be helpful, as debugging such things can be hard.

grembo:
There are many more numeric Settings stored as strings (like timeouts etc), but not all. There are also more things to fix around devfs rules, but I would like to limit the scope of this change, so we can finish it and do a patch in the ports tree - the current behavior can make systems more vulnerable/less resilient and should be fixed ASAP. If you want to change devfs_rules to numeric later, I can change min_... to numeric now.

sonicaj:
@grembo let's please convert this to integer right now ( with devfs_ruleset too ) while we are bumping a config version. Also let's please make sure that if user already has an invalid value configured, we handle that case gracefully and default to 4 like we do for other jails in the correcting migration.

Also let's please ensure that the user is not able to add an invalid value for this i.e string. Please let me know if you run into any issues.

Thank you

grembo:
> @grembo let's please convert this to integer right now ( with devfs_ruleset too ) while we are bumping a config version. Also let's please make sure that if user already has an invalid value configured, we handle that case gracefully and default to 4 like we do for other jails in the correcting migration.

_Update: Not sure if graceful fallback to 4 in case of invalid input is needed, as this already dies in the current version of iocage (1.2) with a ValueError and in previous versions it would fail on jail start due to using an invalid devfs_ruleset id. So all this would do is unbreak setups that never worked before, which can be surprising at best or even dangerous, while papering over new invalid setups that are created by manually editing config.json._

> 
> Also let's please ensure that the user is not able to add an invalid value for this i.e string. Please let me know if you run into any issues.
> 
> Thank you

Do you really want conf['devfs_ruleset'] and conf['min_dyn_devfs_ruleset'] to be ints? It would make them the first ones in the entire configuration (besides truthy_props, which are handled separately and have distinct logic all over the place).

All other numeric values like _exec_timeout_ etc. are all strings. This would also mean a lot of changes to the settings infrastructure in ioc_json.py + changing all places where devfs_ruleset is used (e.g. in ioc_start.py). Lots of room for introducing errors.

Wouldn't it make more sense to keep them as strings, but validate them on set/load, so most of the code can stay untouched? Converting settings to ints seems more like a project to be done separately and applied to all values that are numeric by nature (like timeouts and counts).

The end-goal of this patch was to fix a security problem, as removing existing devfs_rulesets (and then applying a default one) is a bad thing. I would be grateful if we won't widen the scope of this patch further. If it helps, I could also remove the 'min_dyn_devfs_ruleset' configuration, avoiding bumping the revision and returning to a hard coded value in generate_devfs_ruleset for the time being (in the end, having this detail not tunable right now shouldn't be a showstopper).

Like I said before, there are more problems with devfs_rules (e.g. how they're handled while the jail is running - you can't see what is configured, but only what is applied, which has funny side effects), the man page is wrong on them etc. - all these things can be fixed later/separately.

sonicaj (commented):
@grembo we are already doing a config bump and it made sense to have the props in question updated to be proper integers. But yes, you are right that we have quite a number of other props which could use the same treatment which we can indeed retarget for a later date.

Let's please only add validation for the devfs props and have this merged ;)

sonicaj (approved):
LGTM. Thank you @grembo. @skarekrow can you please have a peek as well ? Thank you

skarekrow (approved):
LGTM too, thanks @grembo 👍

grembo:
@skarekrow @sonicaj Cool! Once this has been merged, I'll open a PR with @araujobsd  to patch the FreeBSD port _sysutils/iocage_.

grembo:
Added patch in FreeBSD ports: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=243510
@dlangille Maybe interesting to you

dlangille:
I noticed this today:

```
[dan@r720-01:~] $ sudo iocage stop bacula-sd-02
* Stopping bacula-sd-02_int_unixathome_org
  + Executing prestop OK
  + Stopping services OK
  + Removing devfs_ruleset: 1005 OK
  + Removing jail process OK
  + Executing poststop OK
[dan@r720-01:~] $
```

This differs from the start:

```
[dan@r720-01:/iocage/jails] $ iocsudo iocage start bacula-sd-02
* Starting bacula-sd-02_int_unixathome_org
  + Started OK
  + Using devfs_ruleset: 1005 (cloned from devfs_ruleset 42)
  + Using IP options: ip4.addr=ix0|10.55.0.33 ip4.saddrsel=1 ip4=new ip6.saddrsel=1 ip6=new
  + Starting services OK
  + Executing poststart OK
[dan@r720-01:/iocage/jails] $ 
```

Can the stop also mentioned 'cloned from 42'? I ask so that I don't wonder why there is no 1005 ruleset in the conf file:

```
[dan@r720-01:/iocage/jails] $ grep 1005 /etc/devfs.rules
[dan@r720-01:/iocage/jails] $ 

```

grembo:
@dlangille Please see #1138 - once differentiating between the configured and the dynamically created ruleset works, it will be easy to do what you’re asking for.

PR #1148, opened 2020-03-13
NAS-105387 / 12.0 / Mount/Umount properly during iocage update failure

sonicaj:

- [ ] Explain the feature

### DIFF
--- a/iocage_lib/ioc_fetch.py
+++ b/iocage_lib/ioc_fetch.py
@@ -861,7 +861,6 @@ class IOCFetch:
                 _callback=self.callback,
                 silent=self.silent)
 
-        su.Popen(cmd).communicate()
         shutil.copy("/etc/resolv.conf", f"{mount_root}/etc/resolv.conf")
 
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_fetch.py
```diff
@@ -944,6 +946,7 @@ def fetch_update(self, cli=False, uuid=None):
                         )
 
             finally:
+                su.Popen(['umount', f'{mount_root}/dev']).communicate()
```
william-gr, line 948: Wont this try to umount twice if you hit the exceptions?

sonicaj: No, `EXCEPTION` level means that we will exit this call for sure the first time it occurs, that's why the need to have these on multiple places

william-gr: Are you sure? finally should run even if you raise an exception.

sonicaj: Oh, okay the above one is unintended :) Good catch. I was thinking that we only have 2 umount commands - should've looked properly.

william-gr (approved):

PR #1154, opened 2020-03-27
NAS-105538 / 12.0 / Make sure mac_prefix is valid

sonicaj:
This PR introduces changes to ensure that the auto generated `mac_prefix` is valid in terms of it having the first bit in the first byte unset and the second bit in the first byte set.
Why we do this is because first bit `set` in the first byte represents multicast address from which a bridge is not able to learn and the second bit `set` represents local addresses which a jail is essentially. So we enforce this restriction now to make sure `mac_prefix` adheres to it.

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -580,12 +580,19 @@ class IOCConfiguration:
             # Helps avoid clashes with other systems in the network
             mac_prefix = default_mac[0]['addr'].replace(':', '')[:6]
 
-            return mac_prefix
         except KeyError:
             # They don't have a default gateway, opting for generation of mac
             mac = random.randint(0x00, 0xfffff)
 
-            return f'{mac:06x}'
+            mac_prefix = f'{mac:06x}'
+
+    @staticmethod
+    def validate_mac_prefix(mac_prefix):
+        valid = len(mac_prefix) == 6
+        if valid:
+            binary = format(int(mac_prefix, 16), '024b')
+            valid = binary[7] == '0' and binary[6] == '1'
+        return valid
 
     def json_write(self, data, _file="/config.json", defaults=False):
         """Write a JSON file at the location given with supplied data."""
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -593,6 +593,9 @@ class IOCConfiguration:
         # that this mac is being used in a local network which we set it
         # always.
         if not IOCConfiguration.validate_mac_prefix(mac_prefix):
+            # First and second bits in the first byte will be at
+            # 7th and 6th indexes respectively as networks are
+            # MSB-LTR ordered
             binary = list(format(int(mac_prefix, 16), '024b'))
             binary[6] = '1'
             binary[7] = '0'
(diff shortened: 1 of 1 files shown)


skarekrow:
IIRC @devinteske did these things for a reason with the mac generation. We should get their input as well.

sonicaj:
@skarekrow happy to make sure that we have everything in order ; )

iocage_lib/ioc_json.py
```diff
@@ -2415,6 +2436,17 @@ def json_check_prop(self, key, value, conf, default=False):
...
                         )
+                elif key == 'mac_prefix':
+                    if not self.validate_mac_prefix(value):
+                        iocage_lib.ioc_common.logit(
+                            {
+                                'level': 'EXCEPTION',
+                                'message': 'Please specify a valid mac_prefix'
+                                           'which has the first bit in the'
+                                           'first byte unset and the second'
+                                           'bit set.'
```
william-gr: This is Greek for 99% of the users :P 
Can we simplify and say mac address cannot start with "XX" or "YY" ?

iocage_lib/ioc_json.py
```diff
@@ -2415,6 +2436,20 @@ def json_check_prop(self, key, value, conf, default=False):
...
+                    # Valid letters - 2,6,A,E
+                    if not self.validate_mac_prefix(value):
+                        iocage_lib.ioc_common.logit(
+                            {
+                                'level': 'EXCEPTION',
+                                'message': 'Please specify a valid mac_prefix'
+                                           'which contains any of '
+                                           'the following letters as a '
+                                           'second character from the left '
+                                           'side: "2,6,A,E".'
```
william-gr: @Mrt134 can you think of a better way to say this?

iocage_lib/ioc_json.py
devinteske: How about:

```
Invalid mac_prefix. Must match `?X` where ? can be any valid hex digit (0-9, A-F) and X is one of 2, 6, A or E.
```

iocage_lib/ioc_json.py
devinteske: Suggest:

```
Invalid mac_prefix. Must match `?X` where ? can be any valid hex digit
(0-9, A-F) and X is one of 2, 6, A or E.
```

iocage_lib/ioc_json.py
```diff
@@ -580,12 +580,33 @@ def get_mac_prefix():
...
 
-            return f'{mac:06x}'
+            mac_prefix = f'{mac:06x}'
+
+        # Reason for this change is that the first bit in the first byte of
+        # mac address dictates unicast/multicast address. In case of
+        # multicast address, bridge does not learn from such addresses.
+        # So we make sure that we have it unset and the second bit indicates
+        # that this mac is being used in a local network which we set it
+        # always.
```
devinteske, line 594: Optional: Only thing I would add here is a note about network endian-ness. That networks are MSB-LTR ordered. It speaks to the fact that the mentioned first- and second-bit are at the 6th and 7th indices

iocage_lib/ioc_json.py
sonicaj: @Mrt134 please let me know if you think the above suggestion is good. Thank you.

william-gr (approved):

sonicaj:
Thank you @devinteske

PR #1161, opened 2020-04-15
NAS-105604 / 12.0 / Handle edge case where retrieved mac_prefix can be an interface name

sonicaj:

- [ ] Explain the feature

### DIFF
--- a/iocage_lib/ioc_json.py
+++ b/iocage_lib/ioc_json.py
@@ -31,6 +31,7 @@ import logging
 import os
 import re
 import shutil
+import string
 import subprocess as su
 import sys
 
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_json.py
```diff
@@ -579,6 +580,12 @@ def get_mac_prefix():
             # Use the hosts prefix to start generation from.
             # Helps avoid clashes with other systems in the network
             mac_prefix = default_mac[0]['addr'].replace(':', '')[:6]
+            if len(mac_prefix) != 6 or not set(mac_prefix).issubset(string.hexdigits):
+                # We do this because in certain cases ( very likely due to netifaces not properly
+                # retrieving values ) mac_prefix can be `lo0` which results in an error
+                # as we consider this to be a valid mac prefix below and just try to replace bits
+                # which in this case don't exist resulting in an unintended exception
+                raise KeyError()
```
skarekrow: Can't we raise a nicer error?

skarekrow (changes requested):

iocage_lib/ioc_json.py
sonicaj: Sure we can, thoughts on `ValueError` ?

skarekrow: I think `ValueError` is more appropriate here, but I meant a message explaining what the comment does. They'll hit this and not look in the code for *why*.

sonicaj: So as in the title, this is an edge case and a fault in `netifaces` which we have patched locally. I don't think we should raise an error here, if we are unable to get the mac address of default gateway, that's okay, we can just generate our own. Thoughts ?

skarekrow: Fair enough :)

sonicaj: Perfect ;)

skarekrow (approved):

skarekrow:
Thanks!

sonicaj:
No problem ;)

PR #1164, opened 2020-04-21
Clarify using iocage with VNET and VLANs

### LINKED ISSUE #901: Documentation bug: VNET networking with VLANs
https://iocage.readthedocs.io/en/latest/networking.html#vimage-vnet
> If using VLAN interfaces for the jail host, add the VLAN interface AND parent interface of the VLAN as bridge members.

Sorry, but this is plain wrong. If there are issues that can be "fixed" this way in certain circumstances, please provide an example and I will try my best to help sort it out.

Kind regards,
Patrick

asomers:
The original wording dates from
3f441b72430c1cb9e1fa993f70ea14d6cf9629e2.  I think it refers to putting
the jail host on a VLAN, but not the individual jails.  But this wording
has confused multiple people, since it's also common to put an
individual jail on a VLAN.

Fixes #901

- [ ] Explain the feature

### DIFF
--- a/doc/source/networking.rst
+++ b/doc/source/networking.rst
@@ -164,8 +164,11 @@ Be sure the default gateway knows the route back to the VNET subnets.
 
 **Using VLANs**
 
-If using VLAN interfaces for the jail host, add the VLAN interface AND
-parent interface of the VLAN as bridge members.
+To assign a jail's traffic to a VLAN, add the VLAN interface as a bridge
+member, but not the VLAN's parent.
+
+If using VLAN interfaces for the jail host, on the other hand, add the parent
+as a bridge member, but not the VLAN interface.
 
 .. index:: Configure Network Interfaces
 .. _Configuring Network Interfaces:


asomers:
cc @skarekrow

doc/source/networking.rst
```diff
@@ -164,8 +164,11 @@ Be sure the default gateway knows the route back to the VNET subnets.
 
 **Using VLANs**
 
-If using VLAN interfaces for the jail host, add the VLAN interface AND
-parent interface of the VLAN as bridge members.
+To assign a jail's traffic to a VLAN, add the VLAN interface as a bridge
+member, but not the VLAN's parent.
+
+If using VLAN interfaces for the jail host, on the other hand, add the parent
+as a bridge member, but not the VLAN interface.
```
skarekrow, line 171: It's been a while since I've messed with vlans and jails. This was originally written by @pannon , but I believe both the vlan interface and the parent interface need to be added, not just the parent. But I could be wrong.

asomers: As @nathanjrobertson and myself have both found, adding the parent interface to the bridge is harmful.  The VLANed jail can only pass traffic when the vlan's parent is _not_ in the bridge.

skarekrow:
Thanks!

PR #1182, opened 2020-06-25
NAS-106547 / 12.1 / Clear out dhcp/accept_rtadv entries from rc.conf if DHCP is disabled

sonicaj:

- [ ] Explain the feature

### DIFF
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -1439,14 +1439,19 @@ class IOCStart(object):
                 if addr.upper() == check_var:
                     nics.append(nic)
 
-        with open(os.path.join(self.path, 'root/etc/rc.conf'), 'r') as f:
-            entries = {
-                k: v.replace("'", '').replace('"', '')
-                for k, v in map(
-                    lambda l: [e.strip() for e in l.strip().split('=')],
-                    filter(lambda l: l.strip() and not l.startswith('#'), f.readlines())
-                )
-            }
+        rc_conf_path = os.path.join(self.path, 'root/etc/rc.conf')
+        if not os.path.exists(rc_conf_path):
+            open(rc_conf_path, 'w').close()
+            entries = {}
+        else:
+            with open(rc_conf_path, 'r') as f:
+                entries = {
+                    k: v.replace("'", '').replace('"', '')
+                    for k, v in map(
+                        lambda l: [e.strip() for e in l.strip().split('=')],
+                        filter(lambda l: l.strip() and not l.startswith('#'), f.readlines())
+                    )
+                }
         for nic in nics:
             if 'vnet' in nic:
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_start.py
```diff
@@ -1424,38 +1422,45 @@ def __start_generate_vnet_mac__(self, nic):
 
         return mac_a, mac_b
 
-    def __check_dhcp_or_accept_rtadv__(self, dhcp):
+    def __check_dhcp_or_accept_rtadv__(self, ipv4, enable):
         # legacy behavior to enable it on every NIC
-        if dhcp and self.conf['dhcp']:
+        if ipv4 and (self.conf['dhcp'] or not enable):
```
skarekrow, line 1427: Don't you mean `or enable` here?

sonicaj: Nope, in that case we want the `else` part to be executed because it means that user has custom specified DHCP on one of the entries in `ip4_addr`. This is to handle the case where DHCP is not desired

iocage_lib/ioc_start.py
```diff
@@ -271,14 +271,12 @@ def __start_jail__(self):
             }, _callback=self.callback,
                 silent=self.silent)
 
-        if wants_dhcp:
-            self.__check_dhcp_or_accept_rtadv__(dhcp=True)
+        self.__check_dhcp_or_accept_rtadv__(True, wants_dhcp)
```
skarekrow: My suggestion is to use kwargs here in case you change the interface of that function again.

iocage_lib/ioc_start.py
```diff
@@ -271,14 +271,12 @@ def __start_jail__(self):
...
 
         if rtsold:
             self.__check_rtsold__()
 
-        if 'accept_rtadv' in self.ip6_addr:
-            self.__check_dhcp_or_accept_rtadv__(dhcp=False)
+        self.__check_dhcp_or_accept_rtadv__(False, 'accept_rtadv' in self.ip6_addr)
```
skarekrow: Here too

iocage_lib/ioc_start.py
sonicaj: This is an internal function so even if we do change the signature later on, we should be good ? Please let me know your thoughts on this, thanks

skarekrow: Well if you change the signature later, `True` could be passing True to something completely different. It's more explicit to see `ipv4=True` imo.

sonicaj: Ah like that, of course we can do that. I thought you were referring to the signature, will change. Thanks

skarekrow (dismissed):

william-gr (approved):

sonicaj:
backport

PR #1194, opened 2020-07-15
Do not fetch/install lib32 for i386

astralblue:
`i386` is a 32-bit architecture therefore lacks a separate `lib32` tarball.  `iocage fetch` fails on `i386` platform when it tries to fetch the non-existent `lib32` tarball.

The same check is already performed for `arm64`; extend the `if` clause to check for `i386` as well.

- [x] Explain the feature


skarekrow (approved):
Thanks!

PR #1198, opened 2020-08-01
NAS-106965 / 12.1 / Make sure we don't mark a succcessful package installation as failed

sonicaj:
If a package fails to install, we output subsequent successful package installation as failed whereas in reality they have not failed as the flag is being reused.

### DIFF
--- a/iocage_lib/ioc_create.py
+++ b/iocage_lib/ioc_create.py
@@ -947,6 +947,7 @@ class IOCCreate(object):
 
             pkg_retry = 1
             while True:
+                pkg_err = False
                 cmd = ("/usr/local/sbin/pkg", "install", "-q", "-y", pkg)
 
                 try:


jsegaert (commented):
This corrects the incorrect messaging.  However, for performance reasons, would it make sense to check the `keep_jail_on_failure` flag [here](https://github.com/iocage/iocage/blob/master/iocage_lib/ioc_create.py#L981) and stop installing additional packages if the plugin-jail is going to be destroyed anyway?  

```
(iocage-env) ➜  iocage git:(NAS-106965) ✗ sudo iocage fetch -P ~/qbittorrent.json -n test-qb dhcp=on --branch 11.3-RELEASE
Plugin: qbittorrent
  Official Plugin: False
  Using RELEASE: 11.3-RELEASE
  Using Branch: 11.3-RELEASE
  Post-install Artifact: https://github.com/freenas/iocage-plugin-qbittorrent.git
  These pkgs will be installed:
    - qbittorrent-nox
    - lang/python3

Testing Host DNS response to pkg.FreeBSD.org
Testing test-qb's SRV response to pkg.FreeBSD.org
Testing test-qb's DNSSEC response to pkg.FreeBSD.org

Installing plugin packages:
  - qbittorrent-nox... 
    - qbittorrent-nox failed to install, retry #1
    - qbittorrent-nox failed to install, retry #2
    - qbittorrent-nox failed to install, retry #3
  - lang/python3... 

pkg error:
  - qbittorrent-nox :

Refusing to fetch artifact and run post_install.sh!
test-qb had a failure
Exception: SystemExit Message: 1
Partial plugin destroyed
```

sonicaj:
Thank you for pointing that out @jsegaert, however that would require quite some number of changes due to how the code is structured if we want to do this the right way.
We are happy to accept a PR if you can please make one with that change ; )

william-gr (approved):

PR #1199, opened 2020-08-08
NAS-106984 / 12.1 / Donot set hostname in rc.conf

sonicaj:
When a jail is created, we set the hostname in rc.conf for each jail, however this means that the value specified in host.hostname never gets respected. So we should either be setting the hostname in rc file each time the jail starts or default to using host.hostname. I propose the latter as I think we should not be making decisions on the user's behalf where he/she might override it in rc.conf his/herself.

- [ ] Explain the feature

### DIFF
--- a/iocage_lib/ioc_create.py
+++ b/iocage_lib/ioc_create.py
@@ -1039,10 +1039,6 @@ ipv6_activate_all_interfaces=\"YES\"
             su.Popen(
                 ['mount', '-F', f'{location}/fstab', '-a']).communicate()
 
-        su.Popen(['sysrc', '-f', f'{location}/root/etc/rc.conf',
-                  f'hostname={host_hostname.replace("_", "-")}'],
-                 stdout=su.PIPE).communicate()
-
         if basejail:
             su.Popen(
                 ['umount', '-F', f'{location}/fstab', '-a']).communicate()


skarekrow:
The value does get respected, on creation. We should likely change it each start of the jail too. But I thought we already did...

sonicaj:
It does not. Changing `host.hostname` has no effect as it is overridden by the value being provided in the rc script. I don't think we should be setting hostname in the rc file on each start of jail. Use case is giving a jail to a user for example and he/she can tailor it to his/her needs. If host user restarts the jail, it will reset the user specified hostname value as well. Also `host.hostname` serves no purpose in this case

sonicaj:
Ah, you mention on creation @skarekrow :D Yes it does get respected on creation and afterwards but always. Changing `host.hostname` has no effect. I don't think we should be doing this in the rc file and just setting `host.hostname` should be enough. Please let me know if you think we have a usecase for setting it in the rc file ?

skarekrow:
It’s just the right place to set this. We should just update it whenever the property is changed. I don’t think we should remove it entirely.

sonicaj:
What use case does host.hostname fulfils in this case ?
User expecting to only have jail(8) host.hostname set would have hostname set in rc file

skarekrow:
I can’t recall if jls and such read from the jails rc.conf. I don’t believe they do. But userland inside the jail cannot read the jail property, so we need to set it in the rc.conf (like we currently do) why would you propose removing that?

The only thing we should be doing is ensuring the two lineup at all times. The property and the rc file.

sonicaj:
`jls` does, so whatever the actual hostname of the jail is, that is going to be respected. `host.hostname` is misleading and does not fulfill the purpose at all. I don't understand why you would like to have a property `host.hostname` which is misleading and sets hostname inside rc file ?

sonicaj:
```
root@freenas[~]# iocage create -r 11.3-RELEASE -n check
check successfully created!
root@freenas[~]# iocage start check
* Starting check
  + Started OK
  + Using devfs_ruleset: 5
  + Using IP options: ip4.saddrsel=1 ip4=new ip6.saddrsel=1 ip6=new
  + Starting services OK
  + Executing poststart OK
root@freenas[~]# jls
   JID  IP Address      Hostname                      Path
     1                  check                         /mnt/test/iocage/jails/check/root
root@freenas[~]#
root@freenas[~]# cat /mnt/test/iocage/jails/check/root/etc/rc.conf
cron_flags="$cron_flags -J 15"

# Disable Sendmail by default
sendmail_enable="NO"
sendmail_submit_enable="NO"
sendmail_outbound_enable="NO"
sendmail_msp_queue_enable="NO"

# Run secure syslog
syslogd_flags="-c -ss"

# Enable IPv6
ipv6_activate_all_interfaces="YES"
root@freenas[~]# iocage exec check hostname
check
root@freenas[~]#
```
This is a jail created with my patch

skarekrow:
It’s been many years since I had to implement that :P there would have been a reason, maybe @pannon can recall? If it works without it, great!

sonicaj:
@skarekrow it certainly does. Let me know otherwise ; )

william-gr (approved):

PR #1203, opened 2020-08-25
Fix resolv

dgeo:
This PR just removes a forced blind copy of host's resolv.conf.

This way jail's config is respected (and not overriden without warning on ``iocage update``)


skarekrow:

PR #1204, opened 2020-08-26
NAS-107235 / 12.1 / Error out update in case of malformed freebsd-update.conf

sonicaj:

- [ ] Explain the feature


sonicaj:
@skarekrow would you recall why we do not have this as a fatal error before in update process ? I just want to be sure that it's not intentional and I am not missing a potential case ?

william-gr (approved):

skarekrow:
Nope likely just hadn’t run into it yet.

PR #1205, opened 2020-08-27
NAS-107346 / 12.1 / Improve iocage library speed by efficiently handling cache

sonicaj:

- [ ] Explain the feature

### DIFF
--- a/iocage_lib/ioc_check.py
+++ b/iocage_lib/ioc_check.py
@@ -44,6 +44,7 @@ class IOCCheck(object):
     def __init__(
         self, silent=False, callback=None, migrate=False, reset_cache=False,
     ):
+        self.reset_cache = reset_cache
         if reset_cache:
             cache.reset()
         self.pool = iocage_lib.ioc_json.IOCJson(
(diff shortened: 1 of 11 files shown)


iocage_lib/ioc_check.py
```diff
@@ -56,9 +57,9 @@ def __init__(
         self.__check_fd_mount__()
         self.__check_datasets__()
 
-        self.pool_root_dataset = Dataset(self.pool, cache=False)
+        self.pool_root_dataset = Dataset(self.pool, cache=reset_cache)
```
themylogin, line 60: I don't get it, `cache=True` means "use cached values", `reset_cache=True` means "do not use cached values", doesn't it?

sonicaj: Exactly, so if user said reset cache, we do that when the class is initialised and we don't have to do it each time we create a dataset which basically means that use the cache now as it has been reset already.

themylogin: I see, thank you for explanation.

themylogin (approved):

william-gr (approved):

PR #1217, opened 2020-09-21
fix #1202

### LINKED ISSUE #1202: iocage update overwrites jail's /etc/resolv.conf file
We are using iocage version 1.2 on several hosts.

When issuing `iocage update <jail>` we noticed that the `/etc/resolv.conf` file of the jail is forcefully overwritten with the host's one, even if the `resolver` property of the jail is set to a value different of `none`.

The problem seems to be here in the `fetch_update` function defined in `iocage_lib/ioc_fetch.py` : https://github.com/iocage/iocage/blob/c3198327c892aad959fda1dbad64934f8f6fd63b/iocage_lib/ioc_fetch.py#L875

Maybe this this file copy should be conditioned to a test of the `resolver` property value ?

dgeo:
This PR just removes a forced blind copy of host's resolv.conf.

This way jail's config is respected (and not overriden without warning on iocage update)

This fixes #1202 for us

### DIFF
--- a/iocage_lib/ioc_fetch.py
+++ b/iocage_lib/ioc_fetch.py
@@ -861,8 +861,6 @@ class IOCFetch:
                 _callback=self.callback,
                 silent=self.silent)
 
-        shutil.copy("/etc/resolv.conf", f"{mount_root}/etc/resolv.conf")
-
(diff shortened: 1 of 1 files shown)


skarekrow:
LGTM, @sonicaj?

PR #1189, opened 2020-07-13
Use buffered reads for importing

randomsamples:
Modifies code in ioc_image.py to use buffered io. Previous code used `.read()` which reads the entire image into a Python `bytes` object before sending it to zfs. For large images this can cause the system to run out of memory.

PR #1218, opened 2020-09-27
Use streaming reads to avoid blowing memory

randomsamples:
Modifies code in ioc_image.py to use buffered io. Previous code used .read() which reads the entire image into a Python bytes object before sending it to zfs. For large images this can cause the system to run out of memory.

NOTE: This is a resubmission of [#1189](https://github.com/iocage/iocage/pull/1189) after a force push severed commit history. See previous PR.

@igalic, @sonicaj, @skarekrow, @jakeswenson

### DIFF
--- a/iocage_lib/ioc_image.py
+++ b/iocage_lib/ioc_image.py
@@ -284,12 +284,15 @@ class IOCImage(object):
                         )
                     ], stdin=su.PIPE
                 )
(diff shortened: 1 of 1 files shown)


skarekrow (approved):
LGTM @sonicaj?

igalic (approved):
this looks fairly solid to me

PR #1222, opened 2020-11-04
Support passing data file to be used by post_install.sh

shen390s:

- [x] Explain the feature

When we use iocage to fetch a plugin,sometimes we may need to  configure plugin with some confidential information which is not suitable to be put in plugin repository. This pull request will enable us to put this information into a file and it will be copied to plugin jail which can be used by post_install script.


shen390s:
I have checked the test fail cases. The fail should not be introduced by the code in pull request.

PR #1212, opened 2020-09-15
fix for #1159 (https://github.com/iocage/iocage/issues/1159)

ableischwitz:
- added HTTP_PROXY, HTTP_PROXY_AUTH, HTTP_TIMEOUT to environment
    of freebsd-update command in order to make use of a proxy.

Make sure to follow and check these boxes before submitting a PR! Thank you.

- [X] Explain the feature
- [X] Read [CONTRIBUTING.md](https://github.com/iocage/iocage/blob/master/CONTRIBUTING.md)

PR #1225, opened 2020-11-18
Update ioc_exec.py

HarryR7:
iocage upgrade fails on machines having proxy
The http(s)_proxy environment values were not transmitted by default, at least in csh (FreeBSD)
Please test it on machines with and without proxy.


HarryR7:
A better solution may be to pass su_env =os.environ in ioc_upgrade

igalic:
this is a duplicate of https://github.com/iocage/iocage/pull/1212 which, just like this pull request, could do with a better title

> A better solution may be to pass su_env =os.environ in ioc_upgrade

nope. that's too dangerous.

PR #1232, opened 2020-12-16
iocage_lib: fetch freebsd-update from releng/${version} instead

kevans91:
Fetching from the release/ tag means that we get the as-released
freebsd-update.sh, but with the caveat that we'll never get any
follow-up EN/SA that apply to freebsd-update.

Switch to the releng/ branch of the target release so we can pick up
any post-release advancements. As an added bonus, we don't need to
bother with the .0 concatenation since there is a 1:1 transformation
between X.Y-RELEASE -> releng/X.Y instead of the release/ legacy
naming convention.

This is particularly relevant as I anticipate there may be a freebsd-update EN to fix it not regenerating passwd database bits.

### DIFF
--- a/iocage_lib/ioc_upgrade.py
+++ b/iocage_lib/ioc_upgrade.py
@@ -100,9 +100,9 @@ class IOCUpgrade:
 
         self.__upgrade_check_conf__()
 
-        f_rel = f'{self.new_release.rsplit("-RELEASE")[0]}.0'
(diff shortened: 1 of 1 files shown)


igalic (approved):
this looks good to me
the failing test could be fixed by removing whatever is importing `six`, since this is a python 3 only project

but that should be none of @kevans91's concern

skarekrow:
Agreed, thanks @kevans91!

PR #1248, opened 2021-02-26
Use plugin release during artifact clone

fulder:
This is a small bugfix for an issue where plugins are using the TrueNAS/FreeNAS version during plugin artifact fetching instead of the `release` version specified in the plugin manifest.

- [x] Explain the feature

### DIFF
--- a/iocage_lib/ioc_plugin.py
+++ b/iocage_lib/ioc_plugin.py
@@ -1191,7 +1191,7 @@ fingerprint: {fingerprint}
             )
         else:
             self._clone_repo(
-                self.branch, plugin_conf['artifact'],
+                plugin_conf['release'], plugin_conf['artifact'],
                 f'{path}/plugin', callback=self.callback
             )
 


sonicaj:
@fulder I think it breaks existing implementation where maintainers would expect release version of TN to be followed. Also ideally TN freebsd version and plugin release freebsd version should be in sync, so not much changes. Thoughts ?

tprelog:
> where maintainers would expect release version of TN to be followed

If we expect the plugin to follow, then we can just use the master branch in the artifact repo and not set a release branch.

As a plugin maintainer, I expected the plugin to pull artifacts from the corresponding release branch, and not be change automatically without my control. The way it is now, actually makes it a lot harder to maintain and update plugins and causes breaking changes with any warning. For example when plugin were released on CORE, they used the 12.1-RELEASE branch,
so there is a corresponding 12.1-RELEASE branch in the plugin's artifact repo. Now TrueNAS has switch plugins the 12.2-RELEASE, with out any notice (or I may have missed) being given -- Now the plugin is no longer using the 12.1-RELEASE branch it has been tested with and is instead using the master branch. Please understand, I don't mean to come sounding like ***hole here but I need to ask, what's the point of even having a release branch if the plugin does not stick to it? I mean, I'm in a position now, I can not make changes that I wanted to make for the 12.2-RELEASE, because the 12.1-RELEASE is already tying to use that branch.

Maybe there is at least middle ground (or even a better ground), where a plugin maintainer can provide the correct branch (from the artifact repo) that should be used in the manifest, and if it is not provided, would just default back to `self.branch` as it is  currently set.

I think, in general, allowing an artifact branch to set independent of the release would be extremely helpful, additionally this could provide a way for plugin maintainer to actually test the update process, before just pushing it out.

tprelog:
I'm thinking something like this (where branch is optional)
```
{
  "name": "awesome-plugin",
  "release": "12.2-RELEASE",
  "artifact": "http://github/whatever-plugin.git,
  "branch": "1.0",
  "official": "false".
...

```

fulder:
My reasoning here was: during the plugin installation and after the jail is fetched/setup using the plugin manifest and once we are in a locked plugin environment, why should the plugin artifact depend to the host machine (TN) version? Isn't it more logical to always keep the plugin artifact version in sync with the plugin jail version?

What's problematic with current implementation is, just like @tprelog wrote, not being able to lock down a specific plugin artifact version which makes it automatically update rather than doing it in a controlled way, the plugin versioning example could be a good solution for this case instead of blending it together with a FreeBSD release. Having a plugin switch branches just because of a new TN versions could break it and force the maintainers to quickly fix the problem in a "production" like environment, instead of release support for the new jail version when ready (I know there are some plugins out there not having any other branch than 11.X, it's probably just lucky some of them still work running using latest TN and the plugin `master` branch)

But maybe I'm overthinking things here and misunderstanding the plugin development flow. In your specific case @tprelog wouldn't it help to branch out a new `12.1-RELEASE` branch from master (in order to lock it for older TN versions), then start to work on a supported version for `12.2-RELEASE` on a separate branch (during this time the `master` branch can be broken for the `12.2-RELEASE` until the fixes are done) and once ready push it to `master` and port back to `12.2-RELEASE` only?

@sonicaj if this is a breaking change of a by-design implementation I could close this one again and we could maybe think of an alternative (using a non-optional manifest parameter) solution instead?

tprelog:
> Isn't it more logical to always keep the plugin artifact version in sync with the plugin jail version?

**Yes, please !!!**

> In your specific case @tprelog wouldn't it help to branch out a new `12.1-RELEASE` branch from master (in order to lock it for older TN versions), then start to work on a supported version for `12.2-RELEASE` on a separate branch (during this time the `master` branch can be broken for the `12.2-RELEASE` until the fixes are done) and once ready push it to `master` and port back to `12.2-RELEASE` only?

Yes it would be a lot easier and this it what I would hope to expect in the future but as stated...

> What's problematic with current implementation is, just like @tprelog wrote, not being able to lock down a specific plugin artifact version which makes it automatically update rather than doing it in a controlled way, ... Having a plugin switch branches just because of a new TN versions could break it and force the maintainers to quickly fix the problem in a "production" like environment, instead of release support for the new jail version when ready


---

> t's probably just lucky some of them still work running using latest TN and the plugin `master` branch)

I have some plugins that use the master branch - I typically try to use the master branch whenever possible and this allows the plugin to follow along with TrueNAS release. I only push to a "release" branch when there is a reason to do so, like release specific work-around I need to apply, or to lock in the final, 12.1-RELEASE for example.

I always hit the most obscure issues - for example, I have a certain plugin that uses the following
```
curl -s ${R} | sed -e "1s|^| -c ${constraints}\n|" > "${required}"
```

In the 12.1-RELEASE this does not correctly insert the new line after `${constraints}\n` - So what I can do in one efficient line on 12.2-RELEASE, is a breaking change on 12.1 - I realize this example is going backwards, and typically we are moving forward... but this is an example were I can not more forward because the 12.1 is trying to use the 12.2 artifact <s>repo</s> branch

sonicaj:
> I'm thinking something like this (where branch is optional)
> 
> ```
> {
>   "name": "awesome-plugin",
>   "release": "12.2-RELEASE",
>   "artifact": "http://github/whatever-plugin.git,
>   "branch": "1.0",
>   "official": "false".
> ...
> ```

@tprelog this would really be nice, I agree. Can you please perhaps make a suggestion ticket about it ?

And I would also agree to move forward with this if we were introducing plugins, so for now i think this change looks acceptable to me. I'll just run quickly with the existing plugins we have and see if there are any breakages and we can have this change then. Thank you for laying out your thoughts @fulder and @tprelog :)

tprelog:
I still need to get on Jira and make a suggestion ticket, but I've been been playing around with a similar change locally -- It's not perfect but with an understanding of the nuances, maybe it is...

So the change I made, was to revert (or leave) `line 1194` at the original
```python
                self.branch, plugin_conf['artifact'],
```

and at `line 1105` I added the following
```python
        self.branch = plugin_conf['branch'] if plugin_conf['branch'] else plugin_conf['release']
```

This would take a bit more re-working to implement using this branch for the install. ( we are only making this change in the update function so it still requires using the console to add `--branch` for the installation )

Even without a full implementation, I've still found this change to be really helpful

Using previous pseudo-plugin manifest, the `"branch": "next-release-or-update",` would never be included in the "released" plugin manifest - but could be available as an "advanced" option. (as a maintainer, I just edit the manifest locally for my plugin)

```
{
  "name": "awesome-plugin",
  "release": "12.2-RELEASE",
  "artifact": "http://github/whatever-plugin.git,
  "branch": "next-release-or-update",
  "official": "false",
...
```

In my plugin's artifact repo, I have two branches `12.2-RELEASE` and `next-release-or-update`.

The way it works with the above change, when I install the plugin like normal (TrueNAS UI), it still uses the `12.2-RELEASE` branch ( It still tries to match a branch to the RELEASE and falls back to `master` if one does not exist ) But when the plugin is updated, it uses the `next-release-or-update` branch since I've added it to the manifest

Even though it is not using the branch from the manifest during the install, IMO this is perfect for easily installing `12.2-RELEASE` and testing the update to `next-release-or-update` (something not currently possible)

Taking into consideration we can still use `--branch next-release-or-update` to test a clean install from that branch, I think what might be considered a limitation actually helps, as long as your aware of it ( It's not bug, it's a feature :joy: )

fulder:
@tprelog nice testing and points!

> The way it works with the above change, when I install the plugin like normal (TrueNAS UI), it still uses the 12.2-RELEASE branch ( It still tries to match a branch to the RELEASE and falls back to master if one does not exist ) But when the plugin is updated, it uses the `next-release-or-update` branch since I've added it to the manifest

This is an interesting feature for maintainers but I have some thoughts about this.

If the `release` field would be the base of the cloned plugin artifact branch (what this branch/PR changes) wouldn't this be enough for testing an upgrade in a more controlled manner already? As a maintainer you could then install an older version just by editing the manifest, upgrade the `release` field and test a real upgrade between 2 different versions. 

Also if the `next-release-or-update` would be an always existing branch which is then updated when needed and match previous -> latest upgrade? Wouldn't this be problematic if users are still using a plugin e.g. 2 versions back? Or would this be an unique branch per release locked by the plugin index release version? If it is the latter it could work fine but does this really give any extra features comparing to just match it with the `release` version (and this way lock the plugin version) ?

I was more thinking about the new `branch` manifest field as a way to lock the plugin version and just make it not being hard dependent of the FreeBSD like it is today. This could make it easier to both upgrade the plugins but also support older FreeBSD versions. Let me describe it with an example and how I was thinking it could work instead:

* Plugin X is released with `branch` version `12.1.0` and using manifest `release` field equal to `12.1` (which only determines the jail FreeBSD version)
* During the FreeBSD `12.2` release the plugin breaks and needs to be updated, but as the manifest currently locks both the jail version and plugin version it will not be updated until the maintainer feels ready and does it by updating the manifest
* The maintainer can start to work on a new version e.g. a new `add-support-for-12-2` branch and in order to test upgrade/fresh install simply edit the manifest manually and change both the `release` and `branch` fields.
* Once ready and working the `add-support-for-12-2` can for example be merged to `master` and released into a new `12.2.0` branch/tag. Finally the plugin manifest updated to this branch as well as the 12.2 `release`

If there would be a need of a new plugin upgrade (due to dependency issues or some manually third party software installed in the `post_install` script) a new `12.2.1` version could be released with any migration steps between these two versions without touching the `release` and also in a controlled way.

An extra nice thing about this example is that if there's a new bug in the old `12.1` release (even though EoL the maintainer could now decide to still support it) the maintainer can create a new `12.1.1` branch based on `12.1.0` branch/tag and fix the problem. When it works as intended just upgrade the plugin manifest `branch` and port it back to only the 12.1-RELEASE index branch.

tprelog:
> If the `release` field would be the base of the cloned plugin artifact branch (what this branch/PR changes) wouldn't this be enough for testing an upgrade in a more controlled manner already? As a maintainer you could then install an older version just by editing the manifest, upgrade the `release` field and test a real upgrade between 2 different versions.

It is a good step in the right direction, and will certainly helps for testing one release to the next, but I don't think this handles the situation where a "plugin update" falls within the same release

> Also if the `next-release-or-update` would be an always existing branch which is then updated when needed and match previous -> latest upgrade? Wouldn't this be problematic if users are still using a plugin e.g. 2 versions back?

My initial thought for this quick drop in, was that `next-release-or-update` branch is only added to the manifest locally, by the plugin maintainer. When changes to this branch are final, they would be merged to the `release` branch.

My knowledge of Python is little more than a few Google searches, but it seems this would require more work for our additional `branch` to work in full capacity, For example, our changes are being made only to the update function. But there is nothing to set the correct `branch` when a plugin is initially installed from the TrueNAS UI, as most user would.

>Or would this be an unique branch per release locked by the plugin index release version? If it is the latter it could work fine but does this really give any extra features comparing to just match it with the `release` version (and this way lock the plugin version) ?

>I was more thinking about the new branch manifest field as a way to lock the plugin version and just make it not being hard dependent of the FreeBSD like it is today. This could make it easier to both upgrade the plugins but also support older FreeBSD versions. Let me describe it with an example and how I was thinking it could work instead:

Honestly, I think we basically saying the same thing

IMO, the perfect world accounts for both `release` and `version` in the manifest ( though `branch` may be the better word choice, for this explanation I think `version` reads better ) -- Whatever the word, we _allow these fields to be optional_.

As you stated, the `release` would only determine the jail FreeBSD version. I think using the `release` should be encouraged but not mandatory. If there is not a release in the manifest, then the jail would be created ( and updated ) following the same release as the TrueNAS host. The latter could allow for a simple plugin to basically have a "rolling" release.

When cloning the artifact repo, if a `version` has been included in the manifest, clone from the matching `version` branch. These `version` branches could be used as you described -- When there is **not** a `version` in the manifest, or matching branch is not found, clone from a branch matching the `release` and fall back to `master` or `main` if the `release` branch does not exist.

---

Plugins seem to be designed with the assumption we are installing a service or application that is available from Fresh Ports or some other FreeBSD package repository - However, that is not always the case. It would be nice if this plugin `version` could also shown in the TrueNAS UI when there is not a primary package to track.

---

On a side note ( separate issue ), I was pulling my hair out trying to figure out why the upgrade only worked some of the time until I realized there is also inconsistent behavior between running a `plugin update` and `plugin upgrade` - When you run a `plugin update`, the artifact repo is cloned first, so the latest `pre_update.sh` script is used. However, if you run `plugin upgrade` the artifact repo is not cloned until the end, so the `pre_update.sh` may be missing, or using an outdated version.

fulder:
> It is a good step in the right direction, and will certainly helps for testing one release to the next, but I don't think this handles the situation where a "plugin update" falls within the same release

That's true, the current fix isn't really an "ultimate" solution for all scenarios, I'm just wondering how we should move on forward right now, should this PR be merged as a potential (quick) fix for not being able to lock the plugin version at all right now and add plugin versions in another step/PR? Or should we just wait for/implement the (optional) plugin versioning and keep the current branch as is (it could change again with the new implementation/support, see sections below)? Any thoughts on this @sonicaj (did you manage to find any plugins were this would be problematic and break something?) / @tprelog ?

> My initial thought for this quick drop in, was that next-release-or-update branch is only added to the manifest locally, by the plugin maintainer. When changes to this branch are final, they would be merged to the release branch.

Aha, I see, that would probably be a good update process :+1: 

> As you stated, the release would only determine the jail FreeBSD version. I think using the release should be encouraged but not mandatory. If there is not a release in the manifest, then the jail would be created ( and updated ) following the same release as the TrueNAS host. The latter could allow for a simple plugin to basically have a "rolling" release.

I like and concur with this section/description (you're right we are probably thinking/saying the same thing here :smile:). 

> When cloning the artifact repo, if a version has been included in the manifest, clone from the matching version branch. These version branches could be used as you described -- When there is not a version in the manifest, or matching branch is not found, clone from a branch matching the release and fall back to master or main if the release branch does not exist.

I was actually just writing a question regarding a manifest with only `release` set after reading the previous section :laughing:. Although, doesn't this go against this sentence:  "..the release would only determine the jail FreeBSD version.."? If there would be an optional version param, why should we fall back to use a `release` param for the plugin branch at all (if needed the branch could always be locked/updated for that specific `release`?)? Wouldn't it be better to just clone from `master`/`main` if not specified to make it a bit more consistent? Or are there cases where this could still be useful?   

> Plugins seem to be designed with the assumption we are installing a service or application that is available from Fresh Ports or some other FreeBSD package repository - However, that is not always the case. It would be nice if this plugin version could also shown in the TrueNAS UI when there is not a primary package to track.

I like this idea as well :+1: There are for example plugins running multiple ports, e.g. a service requiring a database, this way we could for example create a plugin version with template: `<PLUGIN_VERSION>-<SERVICE_VERSION>-<DATABASE_VERSION>`: `1.0.0-myService3.4-myDatabase5.2` or something like this. This way one part of the plugin can be updated with faster feedback for the user (e.g. just updating the database version) as well as help any bug debugging for a specific version.

fulder:
> On a side note ( separate issue ), I was pulling my hair out trying to figure out why the upgrade only worked some of the time until I realized there is also inconsistent behavior between running a plugin update and plugin upgrade - When you run a plugin update, the artifact repo is cloned first, so the latest pre_update.sh script is used. However, if you run plugin upgrade the artifact repo is not cloned until the end, so the pre_update.sh may be missing, or using an outdated version.

@tprelog I'm a bit confused about this, are you sure this is the correct behavior? Looking at the (`master`) code it looks like `upgrade` is just updating the plugin index, upgrading the FreeBSD jail version and calling the same `update` function. Which should clone/pull the plugin repo before executing the `pre_update.sh` script. But maybe I'm missing something here?

fulder:
I tried to implement the optional plugin `branch` property in: https://github.com/iocage/iocage/pull/1250, maybe this PR could be replaced by that one instead? What do you think @tprelog / @sonicaj :)?

PR #1250, opened 2021-04-05
Optional plugin branch property

fulder:
This PR adds a new optional `branch` plugin property to the manifest. Setting this property in the plugin manifest will clone/pull the plugin artifact from that branch rather than the TrueNAS FreeBSD version and if not present/set fallback to the plugin `master` branch. This way the plugin version can be locked/updated in a more controlled manner and not depend on the TrueNAS (FreeBSD) version which would otherwise automatically update a plugin during a new release.

Related (Closes?) PR/discussion: https://github.com/iocage/iocage/pull/1248

---

- [x] Explain the feature


Baenwort:
I want to mention this would be very helpful for me trying to move MotionEye from Python 2 to Python 3 without breaking things as I wait for the manifest.json to be updated and then not breaking things once it is until I can update the artefact repo.

It would also enable (combined with the release versioning of the manifest.json) maintaining support for stable/old releases of TrueNAS while still supporting the newest during a period of transition like we went through with 12 to 13.

PR #1252, opened 2021-04-10
Replace kFreeBSD instructions with modern Debian.

trasz:
Provide documentation for using modern Debian or Ubuntu jails. This could be done in addition to the old kFreeBSD instructions, but it looks like kFreeBSD seems to have stalled; the existing docs no longer work anyway.

This is largely based on: https://hackacad.net/post/2021-01-23-create-a-ubuntu-linux-jail-on-freebsd/


skarekrow:
Looks great, thanks!

PR #1253, opened 2021-04-17
Cirrus test fixes

fulder:
Attempt to fix the broken cirrus test check. Changes/fixes include:

* Add `six` to the `iocage_cli` package (this was the main reason tests were not started)
* Bug fix not being able to listing jails with `jid` sort when some jails are down with `None` jid values
* Bug fix for starting jail with no `rc.conf` file and no parent `/etc` folder
* Bug fix for trying to get not yet cached dataset value 
* Ignore start exceptions in `__all__` function. This is a fix for a bug while running `iocage restart ALL` and having an empty jail which didn't start (as expected) but made all the following jails being skipped
* Bump the test FreeBSD version and cirrus version to `12.2`
* Restart jail test fix where tests weren't waiting for all jails to get (sequentially) restarted
* `fstab` test fix having invalid params
* Change the upgrade test to a more hardcoded/simple version (these tests are still not run in the cirrus CI as it takes a lot of time and slows down the tests to about twice the execution time)
* Coverage fix not printing any of the covered files

Note that these fixes solves some of the issues and make the tests green again although there is still some flakiness where jails are started but missing the `/dev` mount which makes the `jexec` command for stopping services fail. See e.g. run: https://cirrus-ci.com/task/5140373442920448 (leaving this for other PRs/issues/fixes in order to not make this grow any further).

----

- [x] Explain the feature

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 freebsd_instance:
-  image_family: freebsd-11-3-snap
+  image_family: freebsd-12-2-snap
 
 iocage_tests_task:
   create_pool_script:
--- a/iocage_lib/ioc_common.py
+++ b/iocage_lib/ioc_common.py
@@ -185,6 +185,8 @@ def ioc_sort(caller, s_type, data=None):
 def get_natural_sortkey(text):
     # attempt to convert str to int to facilitate simplified natural sorting
     # integers will be ranked before alphanumerical values
+    if text is None:
+        return 30, None
     try:
         return 10, int(text)
     except ValueError:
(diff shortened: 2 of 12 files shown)


skarekrow:
Thanks for this @fulder! Why the requirement for six suddenly? They used to run just fine with just python 3 but I haven’t kept up on following cirrus developments.

fulder:
>Why the requirement for six suddenly? They used to run just fine with just python 3 but I haven’t kept up on following cirrus developments.

@skarekrow I just noticed the `six` package is currently used in the [ensure_unicode_str](https://github.com/iocage/iocage/blob/master/iocage_lib/utils.py#L22) function referenced in `fstab` related code. Maybe I've misunderstood something here but shouldn't it be included in the package requirements if used?  If you check some of the currently open PRs and their cirrus test runs they all seems to fail due to not available `six` package, for example this run: https://cirrus-ci.com/task/6446850703097856

skarekrow:
Yes you’re right @fulder it should. These were additions made by @sonicaj. I should ping @araujobsd so we know to get the port adjusted as well.

skarekrow (approved):

PR #1265, opened 2021-06-29
Clarify "jail_zfs_dataset" man page description.

SJChannel:
Explain that multiple datasets may be specified and that they should
be separated by whitespace.


skarekrow:
Thanks!

PR #1266, opened 2021-07-12
Updates for Python 3.8 and pip

jsegaert:

- [x] Explain the feature

Changes:
- The default version of Python is now 3.8; this PR addressed this
- Remove old _--recursive_ references from when libzfs was an embedded repo
- The newer versions of pip no longer support the existing setup.py file and abend in `ERROR: More than one .egg-info directory found in /tmp/blahblah`.  Calling `setup.py` directly in the Makefile still works.  There's some background here: pypa/pip#8201.

Thanks!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,11 @@
 ZPOOL=""
 SERVER=""
-PYTHON?=/usr/local/bin/python3.6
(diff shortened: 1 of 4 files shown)


skarekrow (approved):
Thanks!

PR #1274, opened 2021-12-13
Fix retrieve_cpu_sets() output parsing

overhacked:
Due to a greedy `*` atom in the regex, this only gets the last **single** digit of the available cpusets, so `15` becomes `5`. This breaks validation of any jails' configured cpuset property that might be greater than the last output digit. Changing to a non-greedy `*?` causes the regex to match all of the final digits.

- [x] Explain the feature


overhacked:
Test failure seems unrelated to this change?

dgeo (approved):
Good catch here :+1:

PR #1286, opened 2022-08-12
Respect `resolver` configuration when updating jails

grembo:
When updating a jail using `iocage update jailname`, iocage replaced
jailname's `/etc/resolv.conf` with the one from the jailhost, ignoring
the jail's `resolver` configuration.

This patch makes `fetch_update` use the same logic when generating
`/etc/resolv.conf` that is used when (re)starting a jail.

- [x] Explain the feature

### DIFF
--- a/iocage_lib/ioc_fetch.py
+++ b/iocage_lib/ioc_fetch.py
@@ -842,6 +842,10 @@ class IOCFetch:
                 },
                 _callback=self.callback,
                 silent=self.silent)
+            iocage_lib.ioc_common.generate_resolv(
+                iocage_lib.ioc_json.IOCJson(
+                    mount, cli=False).json_get_value('resolver'),
+                mount_root)
         else:
             cmd = [
                 "mount", "-t", "devfs", "devfs",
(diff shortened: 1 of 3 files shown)


grembo:
iocage_tests CI appears to be defective.

iocage_lib/ioc_fetch.py
```diff
@@ -860,8 +864,7 @@ def fetch_update(self, cli=False, uuid=None):
                 },
                 _callback=self.callback,
                 silent=self.silent)
-
-        shutil.copy("/etc/resolv.conf", f"{mount_root}/etc/resolv.conf")
+            shutil.copy("/etc/resolv.conf", f"{mount_root}/etc/resolv.conf")
```
dgeo, line 867: why not the same here ?

dgeo (approved):
Ok for all here
I'd do the same for !cli, but not sure of what is the use-case for this

iocage_lib/ioc_fetch.py
grembo: It's been a while since I opened this, so my memory is limited. Based on the code, there are these code paths:

1. fetch_release: Get a fresh release, cli=False (where copying in a fresh resolv.conf is fine, I guess)
2. jail is a basejail: cli=False - here probably the config should be respected - I never used basejails with iocage
3. everything else (cloned jails): cli=True, this is what I tested

So, in case 1. is also okay (not sure if there is a config involved at that point), yeah, I guess it could be done on all. Somebody would have to test it though :)

grembo:
@dgeo Does #1217 make this patch obsolete?

dgeo:
Not really, your code is way better than mine here !

PR #1291, opened 2022-12-31
Correctly construct arguments for freebsd-update

aWZHY0yQH81uOYvH:
When calling `freebsd-update`, `--currently-running` and the jail's current release should be separate arguments, instead of in the same argument separated by a space. This only works now because `freebsd-update` doesn't correctly handle spaces in its arguments and splits them by spaces internally. If this is fixed ([bug report](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=268686)), `iocage` will not be able to run it properly.


dgeo (commented):
correct

Issue #1289, opened 2022-10-05
Last Accepted Pull Reqest was Oct 2021


PythonLinks:
It makes this project look dead.  But in fact there were two pull requests in Sept 2022, and a few others in the last year all ignored. 

Is there any way to resurrect this project?  To get someone reviewing pull requests.  I would really like to use it.   Bastille looks more alive, but they are based on a NAT model.  Vutr.com gives multiple IPV6 addresses, and so this project, with multiple public ip addresses,  looks better for that use case.

kmoore134:
From the iX side we no longer actively collaborate on this project. Don't believe other members are active as well though. We've mostly moved on to working directly on Linux native containers / Apps on TrueNAS.

centromere:
@kmoore134 Are you planning on moving away from FreeBSD entirely? :(

kmoore134:
For Apps / Plugin / VM workloads our recommendation is to use TrueNAS SCALE / Linux, it'll provide a much better experience overall.

For basic storage functionality (iSCSI/NFS/SMB) using CORE is perfectly fine :)

PythonLinks:
I am very happy with Bastille BSD.  Here is a comparison with Docker

https://forestwiki.com/bastille-vs-docker

jhgit:
> I am very happy with Bastille BSD. Here is a comparison with Docker
> 
> https://forestwiki.com/bastille-vs-docker

The link is currently https://forestwiki.com/bastillebsd-vs-docker

nathanjrobertson:
See Issue #1251 - I'd say the project is dead, sadly. Quite a number of users, no maintainers.

There's not only a few open PRs, there's actually quite a large difference between the most recent released version and what's in `master` (ie. merged but unreleased updates). Some of these have been pulled in by the FreeBSD package maintainer (ie. cherry picked), but most have not.

igalic:
So far, there's no viable fork.
and no viable drop-in replacement\*

so i think it's time people either start thinking about creating a fork, or migrating to an alternative that is actually alive and well

___________
\*I'm of the co-authors of (lib)ioc, which is also dead.

sporkman:
Oof. I don't know much at all about python, but is there any gruntwork to be done on a project like this? Like testing out pulls, creating releases, etc. that a noob could do? I use IOC all over the place and for the most part it continues to "just work" - it seems like even a fork that just includes fixes for any issues brought on by underlying OS changes would be worthwhile.

[...]

PR #1292, opened 2022-12-31
Pull `freebsd-update` from active branch of freebsd/freebsd-src

aWZHY0yQH81uOYvH:
The [master branch](https://github.com/freebsd/freebsd-src/tree/master) of freebsd-src is obsolete and is no longer receiving commits, so `iocage fetch` is using an old version of `freebsd-update`. They switched to the [main](https://github.com/freebsd/freebsd-src/tree/main) branch for active development. [A previous `iocage` commit](https://github.com/iocage/iocage/commit/23192260ead2303be9576a773d0140d98fc356f2) appears to have partially addressed this but neglected to change the branch name.

Note this is still different from the behavior in [ioc_upgrade.py](https://github.com/iocage/iocage/blob/master/iocage_lib/ioc_upgrade.py#L109) which pulls a release version.

### DIFF
--- a/iocage_lib/ioc_fetch.py
+++ b/iocage_lib/ioc_fetch.py
@@ -892,7 +892,7 @@ class IOCFetch:
         su.Popen(cmd).communicate()
         if self.verify:
             f = "https://raw.githubusercontent.com/freebsd/freebsd-src" \
-                "/master/usr.sbin/freebsd-update/freebsd-update.sh"
+                "/main/usr.sbin/freebsd-update/freebsd-update.sh"
 
             tmp = tempfile.NamedTemporaryFile(delete=False)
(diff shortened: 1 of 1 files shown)


igalic:
not to be a downer, but
https://github.com/iocage/iocage/issues/1289#issuecomment-1268638684

For a while, we tried keeping up with it, and improving on it, over at https://github.com/bsdci/ioc / https://github.com/bsdci/libioc
but we ran out of steam / time.

it might be time to start looking into alternatives, or find people motivated enough to fork the project.

aWZHY0yQH81uOYvH:
Yeah, I saw that... It is a little strange that iX is not supporting this given this is the primary tool for handling jails (a major feature) in one of their active products.

dgeo (approved):
I'd merge this one

PR #1294, opened 2023-09-01
add allow_mount_fdescfs option

dgeo:
Patch adapted from https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270016

Iocage has no property to allow the user to mount the fdescfs filesytem inside a jail. This prevents one from running a Samba 4.16 Domain Controller inside a jail.

This fixes that by adding allow_mount_fdescfs

### DIFF
--- a/iocage.8
+++ b/iocage.8
@@ -1644,6 +1644,18 @@ Default: 0
 .Pp
 Source:
 .Xr jail 8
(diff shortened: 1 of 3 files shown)


skarekrow (approved):
Thanks for this!

PR #1295, opened 2023-10-02
iocage start: fix wrong hexmask for newer freebsd versions.

sueszli:
the `iocage start` command has bit math that's unnecessary on newer freebsd versions.

to keep backwards compatibility a version check was introduced that runs the old code for freebsd <12.X.

### DIFF
--- a/iocage_lib/ioc_start.py
+++ b/iocage_lib/ioc_start.py
@@ -907,7 +907,7 @@ class IOCStart(object):
                 # newer freebsd versions use 'ifconfig -f inet:cidr' and don't need any hex conversion
                 freebsd_version = iocage_lib.ioc_common.checkoutput(['freebsd-version']).split('-')[0].split('.')
                 freebsd_version_major = int(freebsd_version[0])
-                if freebsd_version_major < 12:
+                if freebsd_version_major < 11:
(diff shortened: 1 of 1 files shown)


iocage_lib/ioc_start.py
```diff
@@ -902,20 +902,25 @@ def __start_jail__(self):
...
-                hexmask = addr_split[3].decode()
-                maskcidr = sum([bin(int(hexmask, 16)).count('1')])
-
-                addr = f'{self.ip4_addr}/{maskcidr}'
+                addr = 'ERROR, check jail logs'
+
+                # newer freebsd versions use 'ifconfig -f inet:cidr' and don't need any hex conversion
+                freebsd_version = iocage_lib.ioc_common.checkoutput(['freebsd-version']).split('-')[0].split('.')
+                freebsd_version_major = int(freebsd_version[0])
+                if freebsd_version_major < 12:
```
dgeo: should'n it be < 11 ?

dgeo (changes requested):
should be <11, as FreeBSD 11 can use -f inet:cidr

sueszli:
correct - thank you for updating

sueszli:
any updates?

dgeo (approved):
Thank you

sueszli:
thank you!

